PR #6, opened 2020-02-17
Let the validation function decide whether to continue or abort request

stiiifff:
… processing based on the request, and the presence or absence of credentials.

This is a breaking change compared to the current implementation
The validation function is now defined as `Fn(ServiceRequest, Option<T>) -> O` instead of `Fn(ServiceRequest, T) -> O`.
This allows to cater for cases where e.g. some routes don't require authentication, or some other application-specific authorization logic.

Another approach could be to add a middleware setting to decide whether to abort the request if there is no credentials present in the request.

### DIFF
--- a/actix-web-httpauth/examples/middleware.rs
+++ b/actix-web-httpauth/examples/middleware.rs
@@ -6,7 +6,7 @@ use actix_web_httpauth::middleware::HttpAuthentication;
 
 async fn validator(
     req: ServiceRequest,
-    _credentials: BasicAuth,
+    _credentials: Option<BasicAuth>,
 ) -> Result<ServiceRequest, Error> {
     Ok(req)
 }
--- a/actix-web-httpauth/src/middleware.rs
+++ b/actix-web-httpauth/src/middleware.rs
@@ -15,14 +15,14 @@ use crate::extractors::{basic, bearer, AuthExtractor};
 
 /// Middleware for checking HTTP authentication.
 ///
-/// If there is no `Authorization` header in the request,
-/// this middleware returns an error immediately,
-/// without calling the `F` callback.
+/// The 'F' callback is called with the request and
+/// the parsed credentials (or None if no `Authorization`
+/// header was passed in the request).
 ///
-/// Otherwise, it will pass both the request and
-/// the parsed credentials into it.
 /// In case of successful validation `F` callback
 /// is required to return the `ServiceRequest` back.
+/// If an error is returned, the middleware will abort
+/// request processing and return an error immediately.
 #[derive(Debug, Clone)]
 pub struct HttpAuthentication<T, F>
 where
@@ -35,7 +35,7 @@ where
 impl<T, F, O> HttpAuthentication<T, F>
 where
     T: AuthExtractor,
-    F: Fn(ServiceRequest, T) -> O,
+    F: Fn(ServiceRequest, Option<T>) -> O,
     O: Future<Output = Result<ServiceRequest, Error>>,
 {
     /// Construct `HttpAuthentication` middleware
@@ -51,7 +51,7 @@ where
 
 impl<F, O> HttpAuthentication<basic::BasicAuth, F>
 where
-    F: Fn(ServiceRequest, basic::BasicAuth) -> O,
+    F: Fn(ServiceRequest, Option<basic::BasicAuth>) -> O,
     O: Future<Output = Result<ServiceRequest, Error>>,
 {
     /// Construct `HttpAuthentication` middleware for the HTTP "Basic"
@@ -71,7 +71,7 @@ where
     /// // or to do something else in a async manner.
     /// async fn validator(
     ///     req: ServiceRequest,
-    ///     credentials: BasicAuth,
+    ///     credentials: Option<BasicAuth>,
     /// ) -> Result<ServiceRequest, Error> {
     ///     // All users are great and more than welcome!
     ///     Ok(req)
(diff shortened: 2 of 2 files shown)


Jonathas-Conceicao:
I'm not sure I get the point of this breaking change. If I get it right. The goal of this change is to allow authentication to be set to some routes but not others, if so. Why not just wrap the authentication middle on a specific `Resource` instead of the hole App? This is how I personally handle some routes having authentication while others doesn't.

alexander-irbis:
@Jonathas-Conceicao while that solves the issue, it is not so convenient to make complicated setups just to exclude some routes from the authorization. Another issue is CORS: HttpAuth blocks the OPTIONS request from the browser with "401 Unauthorized" and the browser does not send the auth header. Although maybe there is a better way to resolve this conflict..

GopherJ:
Totally agree @alexander-irbis

stiiifff:
> I'm not sure I get the point of this breaking change. If I get it right. The goal of this change is to allow authentication to be set to some routes but not others, if so. Why not just wrap the authentication middle on a specific `Resource` instead of the hole App? This is how I personally handle some routes having authentication while others doesn't.

My use-case is a GraphQL API (using the juniper crate) where queries do not require auth but mutations do ... and both (queries & mutations) go through a single /graphql endpoint.

GopherJ:
Is there anyone who can review this PR and accept it?

Dowwie:
While I agree with the goal of having more flexible middleware, I disagree with how that is being achieved here.  This seems like a flawed design.  I use the other tools in the toolbox to achieve a finer level of authentication and haven't imposed changes upon the actix architecture to do so.

I respectfully disapprove of this change..

stiiifff:
@Dowwie I’m happy to use a better solution if this is too much of a breaking change. What would be your approach for the use case I mentioned above ?

Dowwie:
@stiiifff I've stayed away from GraphQL for a few reasons, security considerations being one (I've considered security too much already).  It may be that the App/Scope/Resource middleware you are working with isn't meant for GraphQL and are the wrong tools for the job-- you may be running into the ["streetlight effect"](https://en.wikipedia.org/wiki/Streetlight_effect).  

If GraphQL is really such a great level up for request/response flow, maybe it deserves its own new approach to authentication and authorization rather than trying to use that which is suited for RESTlike flow.  Sorry that I can't be more specific but I would have to learn more before sharing more concrete ideas.  In your case, you could write a middleware, using actix-net architecture (as-is), that will distinguish mutations from queries and enforce accordingly.

cetra3:
I don't feel that it is worth introducing greater complexity to the standard basic auth/bearer middleware.  It seems like it's a bit of a niche case and the more common case is that you are always expecting the header or return unauthorised.  Otherwise it is less ergonomic to consume the middleware if it's optional.

Is basic auth the recommended approach for GraphQL?  All the guides look like they use tokens.

stiiifff:
@cetra3 I'm using bearer auth, not basic auth.
Agreed that my use case should be catered for at another level than in the auth middleware.

GopherJ:
@Dowwie Is there a good reason for closing this?

GopherJ:
@cetra3 

> I don't feel that it is worth introducing greater complexity to the standard basic auth/bearer middleware. It seems like it's a bit of a niche case and the more common case is that you are always expecting the header or return unauthorised. Otherwise it is less ergonomic to consume the middleware if it's optional.

So you are saying giving users the right to decide on their own is **introducing complexity**?

Dowwie:
The reason this issue is closed is that there is a viable path forward by using the tools already available.  This is only one instance of many where developers, including me, do not fully understand what to use when.  The immediate reaction is to just change and build towards a solution that works, rather then use what is already available.

Dowwie:
> @cetra3 
> 
> > I don't feel that it is worth introducing greater complexity to the standard basic auth/bearer middleware. It seems like it's a bit of a niche case and the more common case is that you are always expecting the header or return unauthorised. Otherwise it is less ergonomic to consume the middleware if it's optional.
> 
> So you are saying giving users the right to decide on their own is **introducing complexity**?

Try to look at this as one episode of a long story where people try to change rather than learn and use that which is already available.  You can help influence change by showing how the existing tooling doesn't suffice.

GopherJ:
@Dowwie So what if we would like to disable this level of secutiry in some cases? And what's the solution for controling many routes but some of those routes don't even need security? Will we need to wrap middleware one by one for those who needs?

Dowwie:
> @Dowwie So what if we would like to disable this level of secutiry in some cases? And what's the solution for controling many routes but some of those routes don't even need security? Will we need to wrap middleware one by one for those who needs?

I recognize the rattling off of questions as impatience and aggravation.  Understand that we have shared incentives to improve this system.  It doesn't make sense to block useful changes.  I don't think anyone would do so. 

Actix-web and actix-web are really powerful as-is.  They already offer multiple ways to solve many problems.  One approach that I've found useful is to enforce fine grained security at the handler level, using custom extractors and or imperative style authorization using values from a request.

If you don't want handler level security and want controls at a more aggregate level, take the time to study your options using middleware.

Show how these are insufficient and we can improve.

GopherJ:
@Dowwie The main issue is:
```
You are closing changes while having no proposed solution
```
I'm patient enough but still, I'm just an actix user, I'd expect a *solution* for my cases, so in the case that you close a PR while not proposing any solution, it can be frustrating I would say, but anyway this will not cause major problems, I would keep using this PR and rebease for new changes.

mike-lloyd03:
Is it possible to introduce a third middleware option that implements this rather than adding a breaking change to `HttpAuthentication::bearer`? Perhaps something like `HttpAuthentication::bearer_optional` that expects a validator function of `Fn(ServiceRequest, Option<BearerAuth>) -> O`? This would not add any new complexity to the currently implemented methods and would allow the users more flexibility when adding bearer auth to their service.

PR #25, opened 2020-03-17
Revert "Disable coverage for PRs"

JohnTitor:
Reverts actix/actix-extras#4


robjtede:
> It's meaningless since GHA disables secrets if PR comes from forks.

this is no longer the case?

JohnTitor:
codecov-action no longer needs token so we can accept PRs from fork repos: https://github.com/codecov/codecov-action#usage

PR #31, opened 2020-03-23
Fix #10 AuthenticationMiddleware was called already panic

### LINKED ISSUE #10: AuthenticationMiddleware was called already panic
We are observing that our actix-web application occasionally panics under a lot of contention with the message "AuthenticationMiddleware was called already". This seems to come from here https://github.com/actix/actix-web-httpauth/blob/v0.4.0/src/middleware.rs#L184

We have recently migrated to actix-web 2.0 and actix-web-httpauth 0.4. We haven't observed this before the migration.

As this doesn't happen for all requests, there seems to be some race condition either within actix-web or actix-web-httpauth that causes this.

We use bearer token authorization and within the validation function, to validate the token, we just perform an HTTP request with reqwest and depending on the result return Ok or Err.

stiiifff:


Jonathas-Conceicao:
Could you provide an example to when the `Service::poll_ready` from `AuthenticationMiddleware` would be called twice?

I would also like to point out that moving from `Arc<Mutex<S>>` to `Rc<RefCell<S>>` will make the `AuthenticationMiddleware` be `!Send`, but I'm not sure if this wold be a problem.

cetra3:
Do you have a repro of the issue you're trying to solve?  It's a breaking change so unless the issue is quite common, may not be worthwhile making it `!Send`

stiiifff:
Trying to reproduce with one of the examples in the actix-web-httpauth repo but not able to.
I'm gonna investigate why it's panicking in my app.
We can close the PR for now.

PR #35, opened 2020-03-28
Change ttl to u32

### LINKED ISSUE #33: actix-redis uses only u16 as TTL
I am not able to set TTL of actix-redis session to more than 65535 seconds (about 18 hours) bcs it uses just u16 variable:
`pub fn ttl(self, ttl: u16) -> Self`

omikuji:
Hi!

I need a long TTL.

closes #33


JohnTitor:
It breaks the current signature, could you also update changelog?

omikuji:
Thank you for checking.

Updated CHANGES.md. Are there any mistakes?

PR #36, opened 2020-03-28
allow user to set the cookie HttpOnly policy for the redis session

BartWillems:
Signed-off-by: Bart Willems

### DIFF
--- a/actix-redis/CHANGES.md
+++ b/actix-redis/CHANGES.md
@@ -1,5 +1,10 @@
 # Changes
 
+## [Unreleased]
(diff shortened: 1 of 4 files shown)


BartWillems:
Are these code coverage tests blocking?

robjtede:
not really blocking but might be nice to include setting and checking in one of the examples

JohnTitor:
Looks good, could you also update changelog?

BartWillems:
@JohnTitor I've updated the changelog

PR #41, opened 2020-04-05
Clarify how to use Session::set_session()

pfrenssen:
It currently is not clear how to use `Session::set_session()`. There is no documentation, and all tests use the same example data which use a strange, double-quoted string, but it is not clarified that this is intended to be data which is serialized into JSON. 

```
Session::set_session(
    vec![("key".to_string(), "\"value\"".to_string())].into_iter(),
    &mut req,
);
```

For users it might not be clear how to handle data types other than strings.

Let's add some variety in the examples with different data types, make it explicit in the examples that we are expecting to see data serialized by `serde_json::to_string()`, and add some documentation to the function.

### DIFF
--- a/actix-session/src/lib.rs
+++ b/actix-session/src/lib.rs
@@ -180,6 +180,10 @@ impl Session {
         }
     }
 
+    /// Adds the given key-value pairs to the session on the request.
+    ///
+    /// Values that match keys already existing on the session will be overwritten. The values in
+    /// the key-value pairs should be serialized into JSON.
     pub fn set_session(
         data: impl Iterator<Item = (String, String)>,
         req: &mut ServiceRequest,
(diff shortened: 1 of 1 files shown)


actix-session/src/lib.rs
```diff
@@ -180,6 +180,10 @@ impl Session {
         }
     }
 
+    /// Adds the given key-value pairs to the session on the request.
+    ///
+    /// Values that match keys already existing on the session will be overwritten. The values in
+    /// the key-value pairs should be serialized into JSON.
```
robjtede, line 186: Maybe: `Keys and values should already be JSON serialized.` reads a bit clearer.

robjtede:
Can you stick an example into the docstring?

actix-session/src/lib.rs
pfrenssen: Good suggestion. I updated the documentation. Note that only the value needs to be serialized.

PR #44, opened 2020-04-12
Compile the examples during CI

### LINKED ISSUE #43: Compile the examples when running tests
Currently when tests are executed the various examples are not being compiled. Let's make sure to do this so the examples will be continuously built against the current versions of the crates they demonstrate. If any incompatibilities arise in the examples when changes are made to their crates we will be alerted in time.

pfrenssen:
Fixes #43


JohnTitor:
Hm, I think the examples should be checked on CI expect for prost-example.

JohnTitor:
Ah, I mean, the examples of actix-redis and actix-web-httpauth can be checked by `cargo check --examples` since it's common example style. But prost-example is project style so we should add workspace entry, then the example can be caught by `cargo check --bins`.

pfrenssen:
Great suggestion! I didn't know about `cargo check --examples`. I will close this and open a new PR.

PR #48, opened 2020-05-04
Update to actix-web 3.0.0-alpha.1

mocsy:
Update all crates in extras to actix-web 3.0.0-alpha.1.
Also upgraded all other deps, including `time`, except `tokio-util = "0.2"` in `actix-redis`.

### DIFF
--- a/actix-cors/Cargo.toml
+++ b/actix-cors/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "actix-cors"
-version = "0.3.0-alpha.1"
+version = "0.3.0-alpha.2"
 authors = ["Nikolay Kim <fafhrd91@gmail.com>"]
 description = "Cross-origin resource sharing (CORS) for actix-web applications."
 readme = "README.md"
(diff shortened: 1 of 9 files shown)


robjtede:
I'm apprehensive to upgrade these while actix-web is still in the alpha phase.

mocsy:
I tou

> I'm apprehensive to upgrade these while actix-web is still in the alpha phase.

I was kind of under the impression that `actix-web` stayed alpha because the ecosystem is not ready yet.

The current approach seems to be doing alpha releases of everything on top of the new `actix-web` alpha, before `actix-web` leaves alpha

Some already updated, see:

https://crates.io/crates/actix-session/0.4.0-alpha.1
https://crates.io/crates/actix-redis/0.9.0-alpha.1
https://crates.io/crates/actix-http/2.0.0-alpha.2

This PR pushes the rest in this repo to the same state.

robjtede:
Maybe, in the spirit of smaller PRs, this makes sense and keeps the release train rolling. Will defer to @actix/core-contributors.

JohnTitor:
I'd prefer not to update version numbers here. Also, I'd avoid updating patch versions without any strong reasons since it's diff-noisy.

JohnTitor:
I'm going to close this as outdated. Thanks for your contribution anyway!

PR #58, opened 2020-05-19
Bugfix: AuthenticationError should set status_code

adamchalmers:
I noticed that my `AuthenticationError`'s `status_code` field was 401, but when I ran `.as_response_error()` the `ResponseError`'s code was 500 (the default value). It's because impl ResponseError for AuthenticationError doesn't define the `status_code()` trait method. Merely setting the status code in `error_response` isn't enough, it seems. I added a unit test for this case too. The test will fail without my change, and pass with it.

### DIFF
--- a/actix-web-httpauth/CHANGES.md
+++ b/actix-web-httpauth/CHANGES.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased]
   - Update the `base64` dependency to 0.12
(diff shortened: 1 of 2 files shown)


adamchalmers:
Tagging @Jonathas-Conceicao and @JohnTitor for review. I don't know if this would be a major, minor or patch change, because none of the type signatures change, but the implementation does. I guess there could be consumers out there depending on the current behaviour.

JohnTitor:
The change looks good, I'd consider releasing this as 0.4.2 since it's actually bugfix. Could you also update the changelog?

adamchalmers:
Yup, done!

PR #72, opened 2020-07-03
actix-cors: cosmetic fixes

lucab:
This rewords part of the docstrings for better readability,
and introduces additional lints. That includes ensuring that
all public types implement the Debug trait.

### DIFF
--- a/actix-cors/src/lib.rs
+++ b/actix-cors/src/lib.rs
@@ -1,11 +1,16 @@
-#![allow(clippy::borrow_interior_mutable_const, clippy::type_complexity)]
-//! Cross-origin resource sharing (CORS) for Actix applications
+//! Cross-Origin Resource Sharing (CORS) middleware for actix-web.
 //!
-//! CORS middleware could be used with application and with resource.
-//! Cors middleware could be used as parameter for `App::wrap()`,
-//! `Resource::wrap()` or `Scope::wrap()` methods.
+//! This middleware can be applied to both applications and resources.
+//! Once built, [`CorsFactory`](struct.CorsFactory.html) can be used as a
+//! parameter for actix-web `App::wrap()`, `Resource::wrap()` or
+//! `Scope::wrap()` methods.
+//!
+//! This CORS middleware automatically handles `OPTIONS` preflight requests.
 //!
 //! # Example
+//!
+//! In this example a custom CORS middleware is registered for the
+//! "/index.html" endpoint.
 //!
 //! ```rust
 //! use actix_cors::Cors;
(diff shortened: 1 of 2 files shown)


JohnTitor:
Looks good! Could you add an entry about Debug impl to the changelog?

actix-cors/src/lib.rs
```diff
@@ -1,9 +1,11 @@
-#![allow(clippy::borrow_interior_mutable_const, clippy::type_complexity)]
```
JohnTitor, line 1: Do not remove this in this PR.

lucab: I think they aren't actually used anymore (or at least local clippy was not complaining). Are you asking to generally keep this annotation, or to submit a separate PR in order to drop it?
(I've added it back and grouped with other annotations for now)

lucab:
@JohnTitor amended, thanks for the quick review!

actix-cors/src/lib.rs
robjtede: I'm in favor of removing these if they're not needed. Preferring more locally scoped exception if they are needed.

actix-cors/src/lib.rs
```diff
@@ -35,10 +37,14 @@
 //!     Ok(())
 //! }
 //! ```
-//! In this example custom *CORS* middleware get registered for "/index.html"
-//! endpoint.
 //!
-//! Cors middleware automatically handle *OPTIONS* preflight request.
+//! In this example a custom CORS middleware get registered for the
+//! "/index.html" endpoint.
```
robjtede, line 42:
```suggestion
//! In this example a custom CORS middleware is registered for the
//! "/index.html" endpoint.
```

also move above the code block

actix-cors/src/lib.rs
```diff
@@ -35,10 +37,14 @@
...
+
+#![allow(clippy::borrow_interior_mutable_const, clippy::type_complexity)]
+#![deny(missing_docs)]
+#![deny(missing_debug_implementations)]
```
robjtede, line 46: combine attributes

actix-cors/src/lib.rs
robjtede, line 42:
```suggestion
//! In this example a custom CORS middleware is registered for the
//! "/index.html" endpoint.
```

also move above the code block

actix-cors/src/lib.rs
robjtede, line 46: combine attributes

actix-cors/src/lib.rs
lucab: Ack, split that one to https://github.com/actix/actix-extras/pull/75.

lucab: Nevermind, I was wrong, my local clippy tricked me somehow. Forcing a check from scratch, I see that those lints are triggered all over the module. I think it makes sense to keep this as is.

robjtede:
Nice cleanup, thanks :)

PR #75, opened 2020-07-06
actix-cors: drop unused clippy overrides

lucab:
This removes some clippy lints overrides which are not anymore
required.


lucab:
Nevermind, I was wrong, these are still needed.

PR #86, opened 2020-07-19
Use `matches` macro to fix clippy warnings

JohnTitor:


robjtede:
Wouldn't this bump our MSRV to 1.42?

JohnTitor:
Ah exactly, thanks for pointing out! I think it's fine here since the MSRV will be 1.41 anyway and 1.42 is three away from the latest stable (1.45). Hmm but it may be good to keep it together with actix-web, we could just suppress this warning until we find another good reason to bump MSRV. What about? @robjtede

robjtede:
I'm in favor of bumping everything to 1.42 for v3 ecosystem release. (I have wanted to use matches! for something before.)

Feels far enough away from stable to support most folk. Though, it should be said, I don't have much experience with what devs expect from MSRVs.

JohnTitor:
> I have wanted to use matches! for something before.

That's exactly "another good reason" =)

In general, three away is far enough as MSRV from my experience. Okay, then let's bump MSRV here.

PR #89, opened 2020-07-24
[DO NOT MERGE] Check if workflow issue still occurs

JohnTitor:
This is to see if the issue still occurs that the status remains pending forever, do not merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,6 @@
 
 > A collection of additional crates supporting the [actix] and [actix-web] frameworks.
 
-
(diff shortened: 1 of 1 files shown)


JohnTitor:
So, this should have had the issue if it still happened like this: https://github.com/actix/actix-extras/runs/891222882
But it seems it's resolved?
I'd take no action here but monitor workflows for a while.

(btw @robjtede I contacted support and they said "minutes are free" means that we have no limit for CI time.)

PR #97, opened 2020-09-11
lint and readme cleanup

robjtede:
## PR Type
Docs + Lint

## PR Checklist

## Overview
- Addresses clippy warnings.
- Makes unrendered readme formatting much more clear without changing rendered output.

### DIFF
--- a/actix-cors/src/lib.rs
+++ b/actix-cors/src/lib.rs
@@ -590,7 +590,7 @@ impl Inner {
                     AllOrSome::Some(ref allowed_origins) => allowed_origins
                         .get(origin)
                         .map(|_| ())
-                        .ok_or_else(|| CorsError::OriginNotAllowed),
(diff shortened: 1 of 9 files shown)


actix-cors/src/lib.rs
```diff
@@ -1061,6 +1061,7 @@ mod tests {
         );
 
         {
+            #[allow(clippy::needless_collect)]
```
JohnTitor, line 1064: This seems not to suppress the warning, according to the annotation?

JohnTitor (approved):
Overall looks good, macOS failures should be spurious.

actix-cors/src/lib.rs
robjtede: Ahh... thanks. You'd think the "needless collect" annotation should go on the expression where the collection actually happens.

PR #106, opened 2020-09-23
CORS: Take TryInto instead of TryFrom

FallenWarrior2k:
## PR Type

Feature

## PR Checklist

## Overview

This allows usage of custom types that implement TryInto for the affected methods, something that was not possible with TryFrom due to the orphan rule.
As with Into, the standard library provides a blanket implementation of `TryInto<U>` for T if U implements `TryFrom<T>`, making this a fully backwards-compatible change.
It also allows for simplification of internal code in some places, as it permits turning nested function calls into method chains.

### DIFF
--- a/actix-cors/CHANGES.md
+++ b/actix-cors/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Unreleased - 2020-xx-xx
 * Implement `allowed_origin_fn` builder method.
+* Use `TryInto` instead of `TryFrom` where applicable to bypass orphan rule
 
 
 ## 0.3.0 - 2020-09-11
--- a/actix-cors/CHANGES.md
+++ b/actix-cors/CHANGES.md
@@ -2,7 +2,9 @@
 
 ## Unreleased - 2020-xx-xx
 * Implement `allowed_origin_fn` builder method.
-* Use `TryInto` instead of `TryFrom` where applicable to bypass orphan rule
+* Use `TryInto` instead of `TryFrom` where applicable. [#106]
+
+[#106]: https://github.com/actix/actix-extras/pull/106
 
 
 ## 0.3.0 - 2020-09-11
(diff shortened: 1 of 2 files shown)


robjtede:
Can you add a test for the TryFrom backwards-compat case and one for the TryInto case.

robjtede:
> This allows usage of custom types that implement TryInto for the affected methods, something that was not possible with TryFrom due to the orphan rule.

It should be noted that the rebalanced cohereance rules in Rust 1.41 that this actually is possible using a manual TryFrom implementation. For example, this code is allowed:


```rust
#[actix_rt::test]
async fn allowed_header_tryfrom_manual() {
    struct ContentType;

    impl TryFrom<ContentType> for HeaderName {
        type Error = Infallible;

        fn try_from(value: ContentType) -> Result<Self, Self::Error> {
            Ok(HeaderName::from_static("content-type"))
        }
    }

    let _cors = Cors::new().allowed_header(ContentType);
}
```

But as you said, there is a blanket impl in the standard lib that means the capabilities after this change are strictly greater so i'd definitely like to move ahead with this.

actix-cors/CHANGES.md
```diff
@@ -1,6 +1,7 @@
 # Changes
 
 ## Unreleased - 2020-xx-xx
+* Use `TryInto` instead of `TryFrom` where applicable to bypass orphan rule
```
robjtede, line 4:
```suggestion
* Use `TryInto` instead of `TryFrom` where applicable. [#106]


[#106]: https://github.com/actix/actix-extras/pull/106
```

robjtede (approved):

FallenWarrior2k:
>It should be noted that the rebalanced coherence rules in Rust 1.41 means this is actually possible using a manual TryFrom implementation.

My bad, I had forgotten about that as I hadn't actually made use of it yet. I just noticed while browsing the docs that some of the trait bounds seemed overly strict (the only common example requiring a `From` trait bound I can come up with off the top of my head is the `?` operator).
Will update the changelog in a moment.

robjtede:
@FallenWarrior2k  there's conflicts from the other CORS PR, simple changes

PR #108, opened 2020-09-28
add MQTT and AMQP packages

robjtede:
## PR Type
??

## PR Checklist
- [ ] Format code with the latest stable rustfmt

## Overview
Might not be worth the trouble but this is tracking adding back in to actix group the AMQP and MQTT. Code is taken from last release commit in ntex repos where package name is `actix-*`.

Do we even want to support these? If so, would maybe need someone to be "code owner" who is actually interested in using/supporting these protocols?


robjtede:
looking at the stats on crates.io this probably isnt worth it

| actix-amqp | actix-mqtt |
|-|-|
| 2020-09-28_11 29 22-a7e35ba1 | 2020-09-28_11 29 28-95cd81e6 |

Neopallium:
maybe dump the code into another repo `actix-old` or `actix-abandoned` and mark as archived?  If someone cares enough to fix it up and maintain it they can do so there or fork it.  Mainly it would be good for people to be able to find it and not start from scratch.

Issue #112, opened 2020-10-08
provide warning when using * in allow_origin


robjtede:
Or panic, either way guide developer to the correct method for acheiving this `send_wildcard`...

JohnTitor:
Closing as implemented by #112.

PR #114, opened 2020-10-09
Panic on wildcard in Cors builder's allowed_origin()

eupn:
## PR Type

Feature

## PR Checklist

## Overview

Implements #112.

### DIFF
--- a/actix-cors/CHANGES.md
+++ b/actix-cors/CHANGES.md
@@ -1,8 +1,10 @@
 # Changes
 
 ## Unreleased - 2020-xx-xx
+* Disallow `*` in `Cors::allowed_origin` by panicking. [#114].
 
-* Disallow `*` in `Cors::allowed_origin` by panicking. [#112].
+[#114]: https://github.com/actix/actix-extras/pull/114
+
 
 ## 0.4.1 - 2020-10-07
 * Allow closures to be used with `allowed_origin_fn`. [#110]
--- a/actix-cors/CHANGES.md
+++ b/actix-cors/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Unreleased - 2020-xx-xx
 
+* Disallow `*` in `Cors::allowed_origin` by panicking. [#112].
 
 ## 0.4.1 - 2020-10-07
 * Allow closures to be used with `allowed_origin_fn`. [#110]
--- a/actix-cors/src/builder.rs
+++ b/actix-cors/src/builder.rs
@@ -115,10 +115,16 @@ impl Cors {
     /// `allowed_origin_fn` function is set, these functions will be used to determinate
     /// allowed origins.
     ///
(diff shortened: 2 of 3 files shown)


actix-cors/CHANGES.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased - 2020-xx-xx
 
+* Disallow `*` in `Cors::allowed_origin` by panicking. [#112].
```
robjtede, line 5:
```suggestion
* Disallow `*` in `Cors::allowed_origin` by panicking. [#114].

[#114]: https://github.com/actix/actix-extras/pull/114

```

actix-cors/src/builder.rs
```diff
@@ -115,10 +115,19 @@ impl Cors {
     /// `allowed_origin_fn` function is set, these functions will be used to determinate
     /// allowed origins.
     ///
-    /// Builder panics if supplied origin is not valid uri.
+    /// # Panics
+    ///
+    /// * If supplied origin is not valid uri, or
```
robjtede, line 120: Does it? Line 114 seems to suggest otherwise.

actix-cors/src/builder.rs
```diff
@@ -115,10 +115,19 @@ impl Cors {
...
+    /// * If supplied origin is a wildcard (`*`). [`Cors::send_wildcard`] should be used instead.
     ///
     /// [Fetch Standard]: https://fetch.spec.whatwg.org/#origin-header
     pub fn allowed_origin(mut self, origin: &str) -> Cors {
+        assert!(
+            origin != "*",
+            "Wildcard in allowed_origin is not allowed. \
+        Please use send_wildcard() instead."
```
robjtede, line 128:
```suggestion
            "Wildcard in `allowed_origin` is not allowed. Use `send_wildcard`."
```

robjtede (changes requested):
thanks for picking this up :)

just some initial thoughts, will sanity check running it later

actix-cors/src/builder.rs
eupn: Sorry I’m not following. How does it?

robjtede: Ahh, I'd forgotten how this works. So it assigns an error to the error field and then `finish` panics if there is an error from anywhere in the process. So in this way it's not *this* method that panics for invalid URLs. Therefore, it should not be documented here (in the same section as panics generated by this method, at least).

Thoughts on that?

eupn: It was there before this PR. Should I remove it?

robjtede: Oh, sorry. Somehow didn't notice that. We can leave it in this PR but I'll likely move it later (probably with some other stuff).

robjtede (approved):

PR #119, opened 2020-10-17
CORS builder rework

robjtede:
## PR Type
Refactor

## PR Checklist

## Overview
Major changes:
- `CorsFactory` struct is removed
  - Moves the middleware intiailization into new_transform, leveraging InitError instead of panicking
- `impl Default for Cors` is overly-restrictive so you can't accidentally have security flaws when relying on the defaults.
- A `Cors::permissive()` constructor that allows anything.
  - I wanted the name to reflect it being not ideal for production use and the docs state this.
- Adds methods for each property to reset to a permissive state, eg `allow_any_origin` and `expose_any_header`.
  - Rationale is that making it too onerous create a reasonable CORS set up will "encourage" use of `Cors::permissive` too much.
  - Naming bikeshed: could be `allow_all_origins` / `expose_all_headers`.
- Fixes current bug where allowed origin functions are not called if allowed_origins is All
  - Makes the semantics of this field more like "allow exact origins" but I think that's reasonable to expect if you've also defined a vaidate function.
- `CorsError::BadOrigin` is no longer an error because String conversions/comparisons do not occur.

### Notes:
The Transform::InitError system needs to be formalized properly at the actix-web level because this way workers fail to start up, no response is recieved to requests and the process does not crash. It's a big weird and kind of reliant on the developer using `env_logger` to notice the error logs.

I think this should be released as 0.5 as a "beta" period for 1.0.

### DIFF
--- a/actix-cors/CHANGES.md
+++ b/actix-cors/CHANGES.md
@@ -2,10 +2,18 @@
 
 ## Unreleased - 2020-xx-xx
 * Disallow `*` in `Cors::allowed_origin` by panicking. [#114].
-* Hide `CorsMiddleware` from rustdocs. [#118].
+* Hide `CorsMiddleware` from docs. [#118].
+* `CorsFactory` is removed. [#119]
+* The `impl Default` constructor is now overly-restrictive. [#119]
+* Added `Cors::permissive()` constructor that allows anything. [#119]
+* Adds methods for each property to reset to a permissive state. (`allow_any_origin`, `expose_any_header`, etc.) [#119]
+* Errors are now propagated with `Transform::InitError` instead of panicking. [#119]
+* Fixes bug where allowed origin functions are not called if `allowed_origins` is All. [#119]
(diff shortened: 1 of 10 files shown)


JohnTitor (approved):
This looks great to me. The overview in the OP and the comments on the source made my review quite easy, thanks!

> I think this should be released as 0.5 as a "beta" period for 1.0.

Yeah, makes sense. Let's do so in case we overlook something :)

robjtede:
Any thoughts on `allow_any￼￼￼￼￼_X` vs `allow_all_Xs` @JohnTitor ?

JohnTitor:
Hm, it's a difficult decision for me as I'm not an English native. In the Rust context, `all` is preferred to `any` AFAIK but I'm not sure how it is in the HTTP/CORS context. Anyway, we could re-consider/rename them when promoting v0.5 -> v1.0 if someone wants, so either works for me atm :)

PR #120, opened 2020-10-19
cors origin validator function receives origin as first parameter

### LINKED ISSUE #115: tracking cors 1.0 promotion
Should be pretty easy to get actix-cors in a 1.0 state since it's feature complete.

- [x] panic on wildcard origin #114
- [x] clean up docs about when panic occurs
- [x] consider removing CorsFactory #119
- [x] address any other open issues
- [x] is disable_preflight needed?
  - im sure someone out there has a use case
  - could consider adding a deprecation notice in v1 asking anyone with a use case can open an issue
- [x] is send_wildcard needed?
  - we'll leave it but it's not very useful in today's web world
- [x] allowed_origin_fn without allowed_origin bug fixed in #119
- [x] pass origin header down as parameter to allow origin fns (+ optimise origin extraction) [#120]
- [x] should CorsError be public?
  - there is probably a use case somewhere for wanting to manually return these error, in which case having the pre-built ErrorResponse impls could be useful.

robjtede:
## PR Type
Featuer

## PR Checklist

## Overview
Makes most common validate-by-function operations simpler by providing the extracted origin as the first parameter to validator functions. 

Closes #115.

### DIFF
--- a/actix-cors/CHANGES.md
+++ b/actix-cors/CHANGES.md
@@ -6,14 +6,18 @@
 * `CorsFactory` is removed. [#119]
 * The `impl Default` constructor is now overly-restrictive. [#119]
 * Added `Cors::permissive()` constructor that allows anything. [#119]
-* Adds methods for each property to reset to a permissive state. (`allow_any_origin`, `expose_any_header`, etc.) [#119]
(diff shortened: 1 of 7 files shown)


JohnTitor (approved):
Looks good!

PR #126, opened 2020-11-07
Add new method to Session

joshbenaron:
## PR Type

Adding new method to `Session` which returns all session data in `HashMap`

## PR Checklist


JohnTitor:
I haven't figured out what the best signature is but it'd be great if we have some tests for this method and/or an example usage on the doc comment, at least.

Issue #127, opened 2020-11-10
Conflict between HttpAuthentication::bearer and actix-cors


salfonseca:
## Expected Behavior
Be able to use HttpAuthentication::bearer and actix-cors in the same project

## Current Behavior
Hi, I am trying to use HttpAuthentication::bearer in a project that also uses actix-cors, if I use HttpAuthentication::bearer or actix-cors alone it works fine, but if I use both I get an error saying that cors has block the request due to No 'Access-Control-Allow-Origin' header is present on the requested resource. 

## Steps to Reproduce (for bugs)
My code is:
```
#[actix_rt::main]
pub async fn init() -> std::io::Result<()> {
 let port = env::var("PORT").unwrap_or_else(|_e| "8000".to_string());
 let mut socket = "127.0.0.1".to_string();
 socket.push(':');
 socket = socket + &port;
 let db_state = db_mongodb::DB::init().await;

 HttpServer::new(move || {
    App::new()
        .wrap(
            Cors::permissive()
        )
        .data(db_state.clone())
        .service(
            web::scope("/api")
            .wrap(HttpAuthentication::bearer(auth::validator))
            .configure(routes::init_secured)
        )
        .service(
            web::scope("/public")
            .configure(routes::init_public)
        )
 })
 .bind(socket).unwrap()
 .run()
 .await
}

```

## Your Environment
* Rust Version: rustc 1.47.0 (18bf6b4f0 2020-10-07)
* actix-web = "3.2.0"
* actix-rt = "1.1.1"
* actix-web-httpauth = "0.5.0"
* actix-cors = "0.5.1"
* Operating System: Debian GNU/Linux 10 (buster)
* Kernel: Linux 4.19.0-12-amd64
* Architecture: x86-64

robjtede:
I suspect there's an error being thrown in httpauth that should not be.

Valentine-Mario:
I also experienced the same issues. Got the same browser CORS error when trying to consume API on the browser after setting the CORS and HttpAuthentication::bearer

Pytal:
I initially thought that I was having the same issue as well with `actix-cors` and `actix-web-httpauth` not working together, but it turned out that the `No 'Access-Control-Allow-Origin' header is present on the requested resource` error on the client was the result of unhandled errors in my routes.

ctron:
I noticed the merged PR, and a new release ('0.5.3`). Still, I see the same behavior.

Should that problem be fixed now?

[...]

PR #128, opened 2020-11-15
fix httpauth extraction error handling in middleware

robjtede:
## PR Type
Fix

## PR Checklist

## Overview
Middleware errors should be returned in such a way that downstream middleware can still operate on them. This patch solves as much of the issue as it can without a breaking change. Specifically, the "extraction" stage is not handled correctly.

Futher work will need to change the interface (since the service request is moved into the process function) in order to correctly handle the "process" stage.

Fixes part of #127

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,4 +10,5 @@ members = [
 ]
 
 [patch.crates-io]
+actix-cors = { path = "actix-cors" }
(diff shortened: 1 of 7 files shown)


fakeshadow (approved):

robjtede:
released in actix-web-httpauth v0.5.1

Issue #87, opened 2020-07-20
session.purge() doesn't do anything and documentation for clear(), purge(), renew()


jrmiller82:
I'm using actix-session with the default `CookieSession` backend. I'm trying to figure out how to delete a cookie from the user's browser. 

`session.purge()` doesn't change anything about the cookie, and since there's no way to get to the `SessionStatus` enum (it's private), when you then load that session in a different handler, even though it was purged, the new handler thinks the session cookie is still valid.

I can run `session.clear()` followed by `session.renew()` and it wipes out the values, but I can't seem to figure out the point of `purge()`.

If there's an easy way to delete the cookie, I think the docs ought to be updated as well. 

I think the docs ought to be updated anyway as well with examples showing when to use `clear()` versus `purge()`, when you explicitly need to call `renew()`.

I also think calling purge should alter the session cookies expiration time to some day in the past so that the cookie drops from the browser session altogether as well.

qkniep:
I had the same problem. When using `session.purge()` the session cookie is not deleted. I use the default CookieSession with only `max_age` and `signed` being explicitly set. I also call `session.purge()` from some endpoint like `/logout`.

So now my explanation for this is that if you call `session.purge()` not from the path to which the cookie is set (so by default "/") a new (independent) cookie is created instead. See also a relevant SO: https://stackoverflow.com/a/23995984/1426630
My workaround so far was to explicitly set the cookie in the `HttpResponse` with path set to `"/"`.

As a fix I propose that the path of the empty cookie created by `session.purge()` be set to the path the session cookie uses. The matching PR for this is #129.

jrmiller82:
Great to hear it's being worked on!

PR #129, opened 2020-11-18
Fix purge from other paths than root

qkniep:
## PR Type
Bug Fix

## PR Checklist

## Overview
Previously the empty cookie from `session.purge()`, which is created in `remove_cookie()` didn't have its `path` set. Now the `path` of this cookie is set to be the same as the session cookie's to ensure it is actually overwritten.

Related to #87


JohnTitor (approved):
Makes sense, LGTM.

robjtede:
released in actix-session v0.4.1

PR #144, opened 2021-01-08
Update dependencies (Tokio 1.0)

andy128k:
## PR Type

Refactor

## PR Checklist

## Overview

Update dependencies to use crates with Tokio 1.0.

### DIFF
--- a/.github/workflows/msrv.yml
+++ b/.github/workflows/msrv.yml
@@ -13,7 +13,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - 1.42.0
+          - 1.46.0
 
     name: ${{ matrix.version }} - x86_64-unknown-linux-gnu
     runs-on: ubuntu-latest
--- a/actix-cors/CHANGES.md
+++ b/actix-cors/CHANGES.md
@@ -1,7 +1,7 @@
 # Changes
 
 ## Unreleased - 2021-xx-xx
-
+* Minimum supported Rust version (MSRV) is now 1.46.0.
 
 
 ## 0.5.4 - 2020-12-31
--- a/actix-cors/Cargo.toml
+++ b/actix-cors/Cargo.toml
@@ -19,7 +19,7 @@ name = "actix_cors"
 path = "src/lib.rs"
 
 [dependencies]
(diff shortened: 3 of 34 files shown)


robjtede:
Thanks for this. I'm going to keep this PR open until v4 is stable to reduce churn since there are more breaking changes to come.

robjtede:
new MSRV is 1.46 btw

andy128k:
PR is updated to latest versions.

Additionally, `actix-redis` is rewritten to not to use deprecated `actix::actors::resolver`.

izik1:
@andy128k actix-web out of date?

andy128k:
@izik1 It is not. Cargo fetches 4.0.0-beta.3 even if `4.0.0-beta.1` is specified. They are semver-compatible.

andy128k:
`redis_async` was updated to 0.9

robjtede:
Alright, I'm declaring this repo ready for tokio v1 merges. We can do one or two sets of betas for the extras crates.

fakeshadow:
trust-dns is removed for the reason of reduce code bloat and improve compile time. tbh I don't see the need to rewrite it in actix-redis. It's not like it need cache and performance. I would suggest you use `actix_tls::connect::default_connector` service. It would be much simplier and would still open up custom dns resolver if needed.

andy128k:
@fakeshadow I am not sure if I understand your suggestion. `redis_async` does not allow to pass an opened socket to it, but just an address.

fakeshadow:
`actix_tls::connect` is the connector service actix-web use. It has a default dns resolvre and also can extend with a custom resolver if you want to.
You can see this piece of doc on how to use it https://docs.rs/actix-tls/3.0.0-beta.4/actix_tls/connect/fn.default_connector.html.

Basically you can box the return type of this function and use it as redis actor state. Then you would be able to call this service, pass a Connect<String> type to it and get a Connection<String, TcpStream> back where the String is host address and TcpStream is already established one after dns resolve.

robjtede:
We can do the resolver rewrite in separate pr. Resolver is not removed in 0.11 only deprecated.

robjtede (approved):
LGTM, nice work

PR #148, opened 2021-01-13
switch from redis-async to redis-rs

### LINKED ISSUE #30: switch from redis-async to redis-rs
redis-rs supports many good features but redis-async doesn't support them. For example. `PSUBSCRIBE` command.

### LINKED ISSUE #46: actix-redis  connect to the redis address with password
when connecting to redis, how to set the access password of the address? When I connect to the redis address with password, no matter how I set it, the connection fails.

arniu:
## PR Type

Feature

## PR Checklist

- [ ] Format code with the latest stable rustfmt

## Overview

Roadmap:

- [x] Add `redis_cmd!`
- [x] Make `Command` private
- [ ] Switch to `redis-rs`

Resolve #30
Resolve #46

### DIFF
--- a/actix-redis/CHANGES.md
+++ b/actix-redis/CHANGES.md
@@ -1,88 +1,91 @@
 # Changes
 
 ## Unreleased - 2020-xx-xx
-* Implement `std::error::Error` for `Error` [#135]
+
+- Implement `std::error::Error` for `Error` [#135]
+- Add `redis_cmd!` macro to compose Redis commands
 
 ## 0.9.1 - 2020-09-12
-* Enforce minimum redis-async version of 0.6.3 to workaround breaking patch change.
 
+- Enforce minimum redis-async version of 0.6.3 to workaround breaking patch change.
 
 ## 0.9.0 - 2020-09-11
-* Update `actix-web` dependency to 3.0.0.
-* Minimize `futures` dependency.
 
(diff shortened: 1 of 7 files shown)


fakeshadow:
I don't see benefit moving to redis-rs. 
redis-async is pretty light and shares all deps with almost any actix-web app.

arniu:
> I don't see benefit moving to redis-rs.
> redis-async is pretty light and shares all deps with almost any actix-web app.

Yes, `redis-async` is much lighter. But it does NOT support SSL or auth.

fakeshadow:
I don't think that's highly desired feature everyone want and people can implement their own solution for that.

That said I'm not against it for additional feature.

fakeshadow:
Update: 

What this PR claim is not complete true. redis-async's client object could not supporting tls and auth(I did not do a full check on the crate so I could be wrong). But actix-redis does not actually use it's client object. actix-redis start a raw tcp connection with redis server and only use types `redis-async` offer to pack command and unpack response.

Therefore it would very possible to add tls and authentication support on top of the current code. I'm working on tls feature right now and it connect to my local tls enabled redis-server just fine

fakeshadow:
Update2:
Auth is already supported by sending command as Command(resp_array!["AUTH", username, password]). So it's not an issue from the beginning.
This command can be called directly from actix-redis though for a fool proof authentication

PR #161, opened 2021-03-02
allow session-only cookies

cortopy:
## PR Type
Feature

## PR Checklist

## Overview

When neither Expire nor Max-Age are present in the `Set-Cookie` header, the browser interprets the cookie as session-only and will be deleted when the browser is closed.

Currently, `RedisSession` public methods don't allow to have session-only cookies even though it seems ready to handle session-only cookies. The `RedisSession` has an Inner with a `max-age` which is an `Option<Duration>` and there is already logic not to include Max-Age if the duration is none. However, there is no way for users to set the max-age as `None`

This PR fixes that by changing the argument of the `set_max_age` function. 

There would be a breaking change for users that have modified the default value, since they would have to pass an Option

I couldn't find any relevant existing tests for this particular module but if this change looks OK, I'll add them

### DIFF
--- a/actix-redis/CHANGES.md
+++ b/actix-redis/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Unreleased - 2020-xx-xx
 * Implement `std::error::Error` for `Error` [#135]
+* Allow the removal of Max-Age for session-only cookies
 
 ## 0.9.1 - 2020-09-12
 * Enforce minimum redis-async version of 0.6.3 to workaround breaking patch change.
--- a/actix-redis/CHANGES.md
+++ b/actix-redis/CHANGES.md
@@ -2,7 +2,9 @@
 
 ## Unreleased - 2020-xx-xx
 * Implement `std::error::Error` for `Error` [#135]
-* Allow the removal of Max-Age for session-only cookies
+* Allow the removal of Max-Age for session-only cookies. [#161]
+
+[#161]: https://github.com/actix/actix-extras/pull/161
 
 ## 0.9.1 - 2020-09-12
 * Enforce minimum redis-async version of 0.6.3 to workaround breaking patch change.
--- a/actix-redis/src/session.rs
+++ b/actix-redis/src/session.rs
@@ -80,8 +80,8 @@ impl RedisSession {
     }
 
     /// Set custom cookie max-age
-    pub fn cookie_max_age(mut self, max_age: Duration) -> Self {
-        Rc::get_mut(&mut self.0).unwrap().max_age = Some(max_age);
+    pub fn cookie_max_age(mut self, max_age: Option<Duration>) -> Self {
+        Rc::get_mut(&mut self.0).unwrap().max_age = max_age;
         self
     }
 
--- a/actix-redis/src/session.rs
+++ b/actix-redis/src/session.rs
@@ -80,6 +80,7 @@ impl RedisSession {
     }
 
     /// Set custom cookie max-age
+    /// Use `None` for session-only cookies
     pub fn cookie_max_age(mut self, max_age: Option<Duration>) -> Self {
         Rc::get_mut(&mut self.0).unwrap().max_age = max_age;
         self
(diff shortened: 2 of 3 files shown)


robjtede:
Thanks, this change sounds fine. Would ask you add test of cookie response and add changelog and such.

cortopy:
@robjtede Thanks for looking into thing. I've added tests, changes and one comment

CI doesn't pass but it doesn't look it's because my changes

actix-redis/src/session.rs
```diff
@@ -80,8 +80,9 @@ impl RedisSession {
     }
 
     /// Set custom cookie max-age
-    pub fn cookie_max_age(mut self, max_age: Duration) -> Self {
-        Rc::get_mut(&mut self.0).unwrap().max_age = Some(max_age);
+    /// Use `None` for session-only cookies
+    pub fn cookie_max_age(mut self, max_age: Option<Duration>) -> Self {
```
ibraheemdev, line 84: This could be `max_age: impl Into<Option<Duration>>` for a bit nicer api.

ibraheemdev (commented):
Looks good to me, but you might as well fix that one clippy lint.

actix-redis/src/session.rs
robjtede: what can be converted into Duration?

ibraheemdev: Not sure, but I mean for the `impl Into<Option<_>>`, so you can pass a value directly. Eh, it doesn't really matter.

robjtede: oh true, for the option<_> part that makes sense... yeah i prefer that

cortopy: not sure I follow, do you mean:
`pub fn cookie_max_age(mut self, max_age: impl Into<Option<_>>) -> Self {`

what does it do?

robjtede: there's a `impl From<T> for Option<T>` so that T's produce Some(T) and None is None.

robjtede: doing that would make this a nonbreaking change

ibraheemdev: @cortopy It would let someone write:
```rust
cookie_max_age(duration); // we don't have to wrap it in `Some` because `T` implements `Into<Option<T>>`
cookie_max_age(Some(duration));
cookie_max_age((duration));
```

cortopy: Rust never ceases to impress me :) thanks the tip

However, I'm getting a compiler error:

```rust
pub fn cookie_max_age(mut self, max_age: impl Into<Option<_>>) -> Self {
        Rc::get_mut(&mut self.0).unwrap().max_age = max_age;
        self
    }
```

```
the type placeholder `_` is not allowed within types on item signatures
```

https://doc.rust-lang.org/error-index.html#E0121

robjtede: see first commetn

actix-redis/CHANGES.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased - 2020-xx-xx
 * Implement `std::error::Error` for `Error` [#135]
+* Allow the removal of Max-Age for session-only cookies
```
robjtede, line 5:
```suggestion
* Allow the removal of Max-Age for session-only cookies. [#161]

[#161]: https://github.com/actix/actix-extras/pull/161
```

actix-redis/src/session.rs
cortopy: I've pushed the changes but it's not my lucky night with CI. No clippy error but all tests pass in my machine, not sure what the error is about

JohnTitor (approved):
This LGTM!

@robjtede Is there any merge blocker?

robjtede:
Now that it's non-breaking, no.

JohnTitor:
Okay, then let's merge it :) Thanks!

robjtede:
released in actix-redis v0.9.2

PR #164, opened 2021-03-22
fix actix-redis by revert most recent changes

fakeshadow:
## PR Type

Bug Fix

## PR Checklist

## Overview

Revert most changes done to actix-redis by https://github.com/actix/actix-extras/pull/144

Reason:
1.  One big point of compile time improvement is from drop trust-dns and use a more light, simple and widely work dns resolver. Import it blindly again in some actix-* repo would void this effort and result in bloat dep tree once again when end user could expect otherwise.
2.  There is dns resolve service in actix-tls crate. The last thing we want is re-write the same logic everywhere.
3.  There is no point abondon working code in favor of something doing the same thing or could end up more limiting.(Please let me know if there is benefit by using redis client directly. As there is no changelog for this change to explain the intention and benefit)

### DIFF
--- a/actix-redis/Cargo.toml
+++ b/actix-redis/Cargo.toml
@@ -22,8 +22,8 @@ default = ["web"]
 
 # actix-web integration
 web = [
-    "actix-service",
-    "actix-web",
+    "actix-web/cookies",
+    "actix-web/secure-cookies",
     "actix-session/cookie-session",
     "rand",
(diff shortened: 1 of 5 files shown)


fakeshadow:
It also worth to note that Error enum has break changes without changelog.

robjtede (approved):

PR #166, opened 2021-03-22
TLS connection support for redis via feature flag

fakeshadow:
## PR Type

Feature

## PR Checklist

## Overview

Add `native-tls` feature for connecting to a tls enabled redis-server.


actix-redis/Cargo.toml
```diff
@@ -53,6 +54,9 @@ rand = { version = "0.8.0", optional = true }
 serde = { version = "1.0.101", optional = true }
 serde_json = { version = "1.0.40", optional = true }
 
+native-tls-crate = { package = "native-tls", optional = true }
+tokio-native-tls = { version = "0.3", optional = true }
```
robjtede, line 58: tokio-native-tls re-exports it's native-tls version

robjtede (commented):
the feature flag should be strictly additive, eg. a new RedisTlsActor

fakeshadow:
redis has it's own schema for decide wether to use tls and potential secure flag.
It would be best be done with a single actor type that parse the input redis address to a uri and construct connector accordingly.
A new actor type would push this task to session module which is out the scope and compilcate the building.

The current feature flag is used to give a limit support for tls and proper builder type should be introduced later when redis authentication and possible uds support added. From there the actor would have enough information to construct according connector types.

actix-redis/Cargo.toml
robjtede: actually looks like you've used it so no need for this entry

robjtede (approved):

fakeshadow:
In a follow PR https://github.com/actix/actix-extras/pull/167. `actix-redis`'s `native-tls` has been relaxed and can connect to both redis and rediss prefixed redis server.

PR #170, opened 2021-03-23
various session api improvements

### LINKED ISSUE #16: How to get all the data from a session store?
**Is there an api to get all the key-value pairs stored in a session store?**

I haven't found how to do it using the actix-session api:
- https://github.com/actix/actix-web/blob/master/actix-session/src/lib.rs#L128-L219
- https://docs.rs/actix-session/0.3.0/actix_session/struct.Session.html#methods

Looking at the api, the value and keys are stored as strings in the `SessionInner` HashMap and it could be possible to provide a method to get access to this store.

*If this is not currently possible, I would be happy to implement an api to expose this. I would need some guidance on the desired api signature because there could be several approaches.*

robjtede:
## PR Type
Feature

## PR Checklist

## Overview

supercedes #126  
closes #16

### DIFF
--- a/actix-redis/examples/authentication.rs
+++ b/actix-redis/examples/authentication.rs
@@ -53,7 +53,7 @@ async fn login(
     let credentials = credentials.into_inner();
 
     match User::authenticate(credentials) {
-        Ok(user) => session.set("user_id", user.id).unwrap(),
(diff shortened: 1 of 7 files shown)


ibraheemdev (approved):
:+1:

PR #197, opened 2021-09-24
protobuf: Update prost to 0.8 (Fixes CI on nightly)

thalesfragoso:
## PR Type

Other

## PR Checklist

## Overview

Breaking change.

### DIFF
--- a/actix-protobuf/Cargo.toml
+++ b/actix-protobuf/Cargo.toml
@@ -22,7 +22,4 @@ actix-rt = "2"
 actix-web = { version = "4.0.0-beta.8", default_features = false }
 derive_more = "0.99.5"
 futures-util = { version = "0.3.7", default-features = false }
-prost = "0.7"
-
-[dev-dependencies]
-prost-derive = "0.7"
+prost = "0.8"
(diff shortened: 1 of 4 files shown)


actix-protobuf/Cargo.toml
```diff
@@ -22,7 +22,4 @@ actix-rt = "2"
 actix-web = { version = "4.0.0-beta.8", default_features = false }
 derive_more = "0.99.5"
 futures-util = { version = "0.3.7", default-features = false }
-prost = "0.7"
-
-[dev-dependencies]
-prost-derive = "0.7"
+prost = "0.8"
```
robjtede, line 25:
```suggestion
prost = { version = "0.8", default-features = false, features = ["std"] }
```

thalesfragoso: We need `prost-derive` for the tests, and I think we would need `resolver = 2` to be able to have separate features for dev-dependencies, which might increase MSRV (don't quite remember when it was stabilized .-. ).

Actually, I think I have a better idea, use prost without the `prost-derive` feature and depend directly on `prost-derive` in our dev-dependencies.

Which one do you prefer ?

robjtede: > We need prost-derive for the tests, and I think we would need resolver = 2 to be able to have separate features for dev-dependencies

That's true in-repo but the published library does not have this problem. (We could add a CI check for it too.) Just add back the dev dep with both features so that lib user's aren't forced to install it.

thalesfragoso: Hmm, do we even need the `std` feature ?

PR #212, opened 2021-12-11
Rework actix session

LukeMathWalker:
## PR Type
Feature + Refactor

## PR Checklist

## Overview
This PR introduces significant breaking changes for `actix-session` and `actix-redis`.
For `actix-redis`:

-  `RedisSession` is removed to be moved into `actix-session` (renamed and reworked into `RedisActorSessionStore`).
- Updated `redis-async` from `0.8` to `0.12`.

For `actix-session`, this is basically a rewrite. Key changes:

- `SessionMiddleware` now requires to explicitly specify the storage backend for the session state. This makes sure that a storage backend is always present when `SessionMiddleware` it's used. Users no longer have to make sure to `wrap` the backend and the middleware in the right order.
- We introduce a `SessionStore` trait to empower users to define their own custom storage backends for session state. All the logic associated with the session cookie is baked into `SessionMiddleware`, significantly reducing the amount of logic required to support a new backend - i.e. library authors just need to specify how to perform CRUD operations. This is a significant improvement with respect to the previous iteration of this crate - both `CookieSession` and `RedisSession` were previously re-implementing the same cookie-handling logic.
- Three storage backends are implemented out of the box, behind features flags: `RedisActorSessionStore`, `RedisSessionStore` (with TLS support), `CookieSessionStore`.
- Tightened the methods exposed on `Session` to prevent users from being able to corrupt the session state by setting fields to string values that are not valid JSON (i.e. `set_session` is now private).


double-daruma:
Question: With `actix-redis` being reworked would it potentially be in scope to optionally encrypt the session data and hash the keys stored in redis to create some protection from an attacker having read or write access to the redis DB?

LukeMathWalker:
Considering the new structure, that would be out of scope for `actix-redis` and would have to be implemented into `RedisActorSessionStore`/`RedisSessionStore`.
Key hashing is already supported - both stores allow the developer to specify a closure that take as input the session key as input and return the cache key as output.
Application-level encryption for the session state can be easily added to the current implementations in a non-breaking fashion, but I'd refrain from doing it in this PR - it is big enough already 😁

daringordon:
> Question: With `actix-redis` being reworked would it potentially be in scope to optionally encrypt the session data and hash the keys stored in redis to create some protection from an attacker having read or write access to the redis DB?

That's neither applicable to a general-purpose open source library such as this but also not a problem that is generally accommodated.

Dowwie:
> Question: With `actix-redis` being reworked would it potentially be in scope to optionally encrypt the session data and hash the keys stored in redis to create some protection from an attacker having read or write access to the redis DB?

Designing to support this scenario would be several steps over-engineered for a general-purpose library such as this.  Also, if your database(s) are compromised, ephemeral session state will be the least of your problems!

Be that as it may, there is nothing stopping you from en(de)crypting all of your session state as you wish within your handlers.

actix-session/src/storage/interface.rs
```diff
@@ -0,0 +1,67 @@
...
+        ttl: &Duration,
+    ) -> Result<SessionKey, UpdateError>;
+
+    /// Delete a session from the store.
+    async fn delete(&self, session_key: &SessionKey) -> Result<(), anyhow::Error>;
+}
+
+#[derive(thiserror::Error, Debug)]
+/// Possible failures modes for [`SessionStore::load`].
+pub enum LoadError {
```
Dowwie, line 38: Having a series of Error types related to sessions, rather than a single Error type, is over-engineered for a library.  What is the rationale?

Dowwie (changes requested):

Dowwie:
Overall, this PR is a strong improvement to the library.  Well done.

actix-session/src/storage/interface.rs
Dowwie: Note that there was a manual implementation of the Error trait in this library that you replaced by adding `anyhow` and `thiserror` as dependencies.  Error proc macros have a tradeoff of convenience for compilation time.  Let's not add proc macros nor new dependencies where they aren't needed, ok?

LukeMathWalker: > Having a series of Error types related to sessions, rather than a single Error type, is over-engineered for this particular library. Considering there was only an InsertError to begin with, it's understand why you continued with that theme, but we will be fine with consolidating to a single enum going forward. Curious what your thoughts are, though.

I think that having a single error enum for distinct operations with different failure modes is anti-pattern, it leads to confusion on the caller side.

Let's go with a specific example: `delete` can never fail due to a serialization error, because there is no serialization in the logic it has to execute.
If we were to have a single `SessionStoreError` with a `GenericError` and `SerializationError` variant we would have to handle the latter as well. This either leads to unnecessary logic on the caller side or, equally problematic, the caller makes an assumption on the implementation details of the `SessionStore` by slapping an `unreachable!` macro on the `SerializationError` branch. The implementation changes at a later date, it now involves serialization, there is no compilation error and the caller starts seeing panics at runtime.

In the specific case of `SessionStore` it is worth stressing that these operations and errors are not going to be a concern for 99% of application developers using this library - the only people who are going to interact with the `SessionStore` trait are store implementers and us, in the middleware. The more precise the interface, the easier it is for a store implementer to provide a store that behaves predictably because its failures are handled correctly by our middleware.

LukeMathWalker: > Note that there was a manual implementation of the Error trait in this library that you replaced by adding anyhow and thiserror as dependencies. Error proc macros have a tradeoff of convenience for compilation time. Let's not add proc macros nor new dependencies where they aren't needed, ok?

I have added `anyhow` to allow store implementers to upcast whatever error is returned by the specific drivers they are using (e.g. `actix-redis` with `redis-async` or `redis`) into a common representation which we can then handle in our middleware code. `anyhow` is the best solution at the moment within the ecosystem, especially if we consider the need to preserve the error cause chain.
I'd be happy to consider alternatives, but I think the addition pulls its own weight in terms of ergonomics for store implementers. `anyhow` does not bring any transitive dependency, it is quite lightweight - I can run some compilation benchmarks if needed.

Re: `thiserror` - we could replace indeed it with manual implementations of `Display` and `std::error::Error`.
These are the dependencies brought in by `thiserror`:
```
── thiserror v1.0.30
│   └── thiserror-impl v1.0.30 (proc-macro)
│       ├── proc-macro2 v1.0.34 (*)
│       ├── quote v1.0.10 (*)
│       └── syn v1.0.82 (*)
```
These three crates are already in our dependency tree, even when no optional features are enabled. They are used by `serde` (derive), by `actix-web` (via derive-more, cookie, etc.), by `actix-server`, etc. - you can see it by launching `cargo tree -i syn` locally. 
So the only added "weight" in terms of dependency is the code in `thiserror` itself. Happy to remove it if you feel strongly about it, but the argument for bloat has nothing to do with proc-macros here.

robjtede: RE: error libs:
- We've gotten by with `Box<dyn StdError>` in -web, would this not do here also vs adding `anyhow`?
- I would prefer using `derive_more` macros for Error impls.

robjtede:
Just wondering what's the choice for the changelog version numbers. Are you expecting that there would be a non-beta release for -session and -redis before this is merged?

LukeMathWalker:
> Just wondering what's the choice for the changelog version numbers. Are you expecting that there would be a non-beta release for -session and -redis before this is merged?

No, I am not. I just consider these changes to be so extensive that I believe they deserve a minor version bump - I believe they go beyond the "we do not guarantee betas to be compatible" policy.

actix-session/src/storage/interface.rs
LukeMathWalker: > I would prefer using derive_more macros for Error impls.

I swapped `thiserror` for `derive_more` in 4a5fe0ee6a4e15e11571243d6842c7f67f773892

LukeMathWalker: > We've gotten by with Box<dyn StdError> in -web, would this not do here also vs adding anyhow?

There are two main pros in favour of using `anyhow::Error` instead of `Box<dyn std::error::Error>`:
- backtraces are always available (if enabled), even on stable;
- the `Debug` implementation traverses the entire error source chain, providing the necessary context to interpret what went wrong. 

The main con is that it requires the underlying error type to be `Send + Sync`.

norru:
Hi, happy to see that there is so much activity on redis-session!

Question: does this change addresses:

- atomic (optimistic concurrency) for Redis session object value updates?
- Redis AUTH
- Redis TLS
- Redis cluster

or, is there a roadmap for those? They seem to be implemented in redis-rs but from what I see in the current version of Actix-web they don't seem to be exposed.

To clarify, I am evaluating at using AWS ElastiCache + actix-web + (potentially) actix-session for a project and at the moment those (missing?) features could be showstoppers.

LukeMathWalker:
Hey @norru!

This PR brings support for TLS and authentication via the `redis-rs` backend.
Cluster support should be possible as well, but I haven't had a chance to dive into it yet.

Can you elaborate on what you mean in your first point?

norru:
1. Redis client A creates session X
1. Redis client B loads session X
1. Redis client A changes session X
1. Redis client A saves session X
1. Redis client B changes session X
1. Redis client B saves session X - changes (3) from redis client A are overwritten without warning. With optimistic locking (versions), this operation would fail and then
1. Redis client B reloads session X
1. Redis client B re-applies changes to session X
1. Redis client B saves session X

Or is this something that is deemed impossible? I can see this scenario happening if client A and client B are web nodes behind a non-sticky load balancer.

LukeMathWalker:
It's not that simple to determine what should happen in step 8.
The changes to the session state are driven by the application code and decisions might have been made based on the state that was retrieved in step 2., which is different from the state that was saved in step 6. - it is not guaranteed that "merging" the changes actually leads to a better outcome/a session state that actually makes sense.

norru:
To be fair, off the top of my head I can't think of any. Just looked Tomcat up and no-one mentions this. In the recent past I've just implemented session management from scratch and designed around it so this problem wouldn't occur (ie sessions were immutable once created, all mutable state was somewhere else).

Maybe I'm just overthinking this, but in the general case I'd like at least to have a callback saying "hey, you're trying to save a session which someone else has changed, here's the current state, here's the state you want to save, let me know what you want to do with it" - which would default for "just stomp on the old values with the new ones".

Dowwie (approved):

LukeMathWalker:
What do we want to do with this one? Are we waiting for the release of a stable v4? @robjtede

robjtede:
Yeah I think to ease transition i'll cut a release without this then merge it immediately after.

robjtede:
0.6 and 0.10 versions of actix-session and actix-redis just released i can finally take a serious look at this and publish in the near term

actix-session/src/storage/redis_actor.rs
```diff
@@ -0,0 +1,294 @@
...
+        let cache_key = (self.configuration.cache_keygen)(session_key.as_ref());
+
+        let res = self
+            .addr
+            .send(Command(resp_array!["DEL", cache_key]))
+            .await?;
+
+        match res {
+            // Redis returns the number of deleted records
+            Ok(RespValue::Integer(x)) if x > 0 => Ok(()),
```
robjtede, line 251: Doing this check is a difference in behavior between this and redis_rs impl. I can't think of a codepath that would get here from a high level; I think you'd have to forge a session. But, for example, this test fails:

```rust
    #[actix_web::test]
    async fn remove_non_existent_session() {
        let store = redis_actor_store();
        let session_key = generate_session_key();
        store.delete(&session_key).await.unwrap();
    }
```

LukeMathWalker: The only case where this might happen is if the session expires in Redis while a request is being handled. I think we can remove it.

LukeMathWalker:
It should be enough to enable `tracing`'s `log` feature flag (see [docs](https://docs.rs/tracing/latest/tracing/#emitting-log-records)) to ensure compatibility with users of `log` while providing higher quality data for those who opt to use `tracing`.

robjtede:
That seems like a really elegant path to upgrading and good to start here. We should enable the `log` feature.

LukeMathWalker:
Done in [c43c2ee](https://github.com/actix/actix-extras/pull/212/commits/c43c2ee65adad949ca64d7beb089c2c2c6aa2317)

LukeMathWalker:
What do you want to do about the outstanding comments? @robjtede

actix-session/src/storage/redis_actor.rs
robjtede: Yes please. I think they should be the same behaviorally from the session perspective.

LukeMathWalker: Done in [fd23949](https://github.com/actix/actix-extras/pull/212/commits/fd2394933f38edb06b885c0d2a0b8fbc7d34d52e)

Dowwie:
> 

Not sure whether you were referring to my comments, but I'm on the same page as you about giving the more granular error types a try.  My thinking continues to evolve.  We'll find out what works best over time.

robjtede:
I'm happy with the more granular error handling.

robjtede (approved):

PR #194, opened 2021-09-11
HttpAuthentication: fix error handling

aliemjay:
## PR Type

Bug Fix

## PR Checklist

- [x] Tests for the changes have been added / updated.
- [ ] Documentation comments have been added / updated.
- [x] A changelog entry has been made for the appropriate packages.
- [x] Format code with the latest stable rustfmt.

## Overview

Closes #90 
Closes #127

PR #213, opened 2021-12-13
stop cloning request across service call

robjtede:
## PR Type

fix

## PR Checklist

## Overview
closes actix/actix-web#2508

### DIFF
--- a/actix-cors/CHANGES.md
+++ b/actix-cors/CHANGES.md
@@ -1,6 +1,9 @@
 # Changes
 
 ## Unreleased - 2021-xx-xx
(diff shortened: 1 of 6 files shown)


thalesfragoso:
This seems very subtle to be a panic, `ServiceRequest` (and thus `HttpRequest`) is readily available for users and I don't see this documented anywhere. We don't need to go far to find mistakes with this "invariant", #194 itself has problems with this (although there should be relative easy to fix).

Using `Rc::get_mut().unwrap()` on a `Rc` that the user can clone anytime doesn't seem very nice.

PR #215, opened 2021-12-16
add test for middleware usage in app and scope

simon-an:
Might help for documenting the usage of this crate.


aliemjay (approved):
LGTM. Thanks you!

PR #220, opened 2022-01-28
fix api documentation link

joshrotenberg:
## PR Type

Documentation 

## Overview

Remove errant `*` that was causing the API Documentation to not be linked.


robjtede (approved):
thanks 👍🏻

PR #233, opened 2022-03-21
Add a new configuration parameter to refresh the TTL of the session even if unchanged

### LINKED ISSUE #231: actix-session 0.6.0: regression. Auto-prolonged sessions are not available anymore
## Expected Behavior

Previous versions of actix-session allowed to create auto-prolonged sessions by specifying `expires_in`.

## Current Behavior

Current implementation allows fixed-length session only (`SessionLength::Predetermined`).

## Context

Typical usage of session is to store some kind of user id, so session is not changed much. Auto-prolonged session is common scenario when users are kept logged in till they interact with a website.

LukeMathWalker:
## PR Type
Feature / Refactoring

## PR Checklist

## Overview
In order to support long-lived sessions with auto-refresh, we need to introduce a new concept into `actix-session`: `TtlExtensionPolicy`.  
This cannot be done in a backwards-compatible fashion because `SessionLength` exposes its fields directly therefore adding new fields to its variants it's a breaking change.

This PR does the following:

- Rename `SessionLength::Predetermined` to `SessionLength::PersistentSession`;
- Refactor `SessionLength` to make it possible to evolve configuration options for its variants, in the future, in a backwards-compatible fashion. Each variant became its own type (`BrowserSession`, `PersistentSession`) and their fields are exposed via builder methods, not directly;
- Rename `SessionLength` to `SessionLifecycle` to use a terminology that is less ambiguous now that we allow the extension of the session length at various points of the session lifecycle;
- Rename `max_session_length` to `session_ttl` in `PersistentSession`;
- Rename `state_ttl` to `session_state_ttl` in `BrowserSession`;
- Add `TtlExtensionPolicy` and a corresponding method on `PersistentSession` and `BrowserSession`. `TtlExtensionPolicy` is marked as `non_exhaustive` to give us the possibility to introduce different strategies in the future without making a breaking change (e.g. refresh the TTL at most once every X minutes);
- Add a new `update_ttl` method to `SessionStore`. This opens up the door to an optimised implementation, depending on the backend, compared to invoking `SessionStore::update` when we simply want to extend the state TTL;
- Move all types related to the configuration of `SessionMiddleware` into a public `configuration` sub-module. This results into a clean homepage for docs, highlighting the types we want users to focus on (`SessionMiddleware`, `Session`).

The default behaviour remains unchanged - we refresh the TTL exclusively on session state changes.

Closes #231.


LukeMathWalker:
This is indeed semver major.
Given that this requires a breaking change, we can leverage the opportunity to provide a dedicated method on `SessionStore` as well. 
I'll add a few more commits later.

actix-session/src/middleware.rs
```diff
@@ -462,52 +489,66 @@ where
...
                                 session_key,
                                 &configuration.cookie,
                             )
                             .map_err(e500)?;
                         }
-
                         SessionStatus::Unchanged => {
-                            // Nothing to do; we avoid the unnecessary call to the storage.
+                            if configuration.refresh_when_active {
+                                dbg!("Here");
```
andy128k, line 522: Hi here!

actix-session/src/middleware.rs
```diff
@@ -181,6 +183,25 @@ pub enum SessionLength {
...
+        ///
+        /// # Performance impact
+        ///
+        /// Refreshing the expiration on activity is not free.
+        /// It is not enough to refresh the session cookie expiration, we need to refresh the TTL
+        /// on the session state as well. This translates into a request over the network if you
+        /// are using a remote system as storage backend (e.g. Redis).
+        /// This impacts both the total load on your storage backend (i.e. number of
+        /// queries it has to handle) and the latency of the requests served by your server.
+        refresh_session_ttl_when_active: Option<bool>,
```
andy128k, line 204: Why not just `bool`?

andy128k: Actually, bool is not very descriptive, and enum would be much better. Either an enum to describe auto-refresh option or a third option in the `SessionLength` enum.

LukeMathWalker: I don't think that a third enum variant in `SessionLength` is the right approach - this only makes sense for persistent session tokens, which is why I nested into this variant.

Using an `Option` is a way to give the user the opportunity to set only some parameters within the variant while saying "I trust you for the rest". I can probably make that more explicit by having a separate type that gets nested inside the variant rather than its fields inlined.

I agree we could use an enum instead of a boolean.

actix-session/src/middleware.rs
LukeMathWalker: Ups 😂

andy128k:
BTW, I noticed that `actix-session` uses `time::Duration` but does not re-exports it.

LukeMathWalker:
Good catch. We can switch to using the one re-exported by actix-web.

actix-session/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "actix-session"
-version = "0.6.1"
+version = "0.7.0-alpha.0"
```
robjtede, line 3: please dont do this it will screw up my release scripts :)

LukeMathWalker: Amended 😛

robjtede: context: https://github.com/actix/actix-web/blob/master/scripts/bump#L34

LukeMathWalker:
Let's get #236 and #235 in before moving forward with this one.

LukeMathWalker:
This is ready to be reviewed now @robjtede 
If you are happy to merge it, I'd suggest to hold off on releasing it. I want to make sure we can support all the functionality in `actix-identity` without having to make further breaking changes.

actix-session/src/storage/redis_rs.rs
```diff
@@ -273,12 +289,11 @@ impl RedisSessionStore {
...
 mod test {
-    use std::collections::HashMap;
-
-    use redis::AsyncCommands;
-
     use super::*;
     use crate::test_helpers::acceptance_test_suite;
+    use actix_web::cookie::time;
+    use redis::AsyncCommands;
+    use std::collections::HashMap;
```
robjtede, line 296: do you have rustfmt re-ordering of imports overriden :P

LukeMathWalker: I am running with the default config of rustfmt, nothing custom.

robjtede: On my machine, neither stable nor nightly re-order the imports automatically like this.

andy128k: Separating empty lines are needed to keep groups of imports.

LukeMathWalker: Oh, that might be it 😅

LukeMathWalker: Opened #237 to sort this with automation in the future.

robjtede (approved):

actix-session/src/config.rs
```diff
@@ -0,0 +1,371 @@
...
+
+use actix_web::cookie::{time::Duration, Key, SameSite};
+
+use crate::{storage::SessionStore, SessionMiddleware};
+
+/// Determines what type of session cookie should be used and how its lifecycle should be managed.
+///
+/// Used by [`SessionMiddlewareBuilder::session_lifecycle`].
+#[derive(Clone, Debug)]
+pub enum SessionLifecycle {
```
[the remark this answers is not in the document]
LukeMathWalker: None that comes to mind, although I don't foresee the need to add new variants in the future. Nonetheless, better safe than sorry perhaps.

robjtede (approved):

PR #234, opened 2022-03-22
add support for session access from guards

tglman:
## PR Type
Feature

## PR Checklist

- [ x] Tests for the changes have been added / updated.
- [ x] Documentation comments have been added / updated.
- [ x] A changelog entry has been made for the appropriate packages.
- [ x] Format code with the latest stable rustfmt.

## Overview
Add support for session access to the `GuardContext` so is possible to do:
```
.guard(guard::fn_guard(|g| {
                    g.get_session() //......
          }))
```


tglman:
Hi, 

I forced push a fix for the failed tests, but the build did not restart automatically, could you re-run the build.

Regards

actix-session/src/lib.rs
```diff
@@ -332,6 +332,13 @@ pub mod test_helpers {
                     .service(resource("/do_something").route(post().to(do_something)))
                     .service(resource("/login").route(post().to(login)))
                     .service(resource("/logout").route(post().to(logout)))
+                    .service(
+                        web::scope("/protected")
```
LukeMathWalker, line 336: This test is already fairly complicated as it is, considering the complexity of the scenario we are going through - I'd prefer not to add more stuff to it.
Can we write separate tests that exercises the guard method? One for the "happy case" (there is a session, we can extract it successfully), one for the unhappy case (there is no session cookie, the request is rejected/fails).

robjtede (approved):

robjtede:
released with v0.6.2

tglman:
Thanks, for the quick review and release.

PR #235, opened 2022-03-23
[actix-session/redis-rs] Catch server-driven disconnections and prevent a user-visible issue via a retry

### LINKED ISSUE #230: actix-session 0.6.0: pipe broken when using redis backend with redis db that has a timeout
When using a redis db that has a timeout set, say for like 5 minutes, if that time period elapses without a request coming in, the connection to the redis db will get disconnected, but actix-session won't reestablish the connection. When a request comes in after that disconnection occurs, the response will be "pipe broken". The actix logging middleware won't even log that a request came in. 

This is a serious issue that took a few days to find the source of the problem. This is easily solvable with a heartbeat feature to stop connections from being broken due to the timeout. 

I used the "redis-rs-session" and "redis-rs-tls-session" features in 0.6.0.

LukeMathWalker:
Fixes #230 

## PR Type
Bug fix

## PR Checklist

This has been tested against the reproducible example provided in #230 - broken before the changes, fixed after.

### DIFF
--- a/actix-session/CHANGES.md
+++ b/actix-session/CHANGES.md
@@ -4,11 +4,11 @@
 - Implement `SessionExt` for `GuardContext`. [#234]
 - `RedisSessionStore` will prevent connection timeouts from causing user-visible errors. [#235]
 - Do not leak internal implementation details to callers when errors occur. [#236]
- 
+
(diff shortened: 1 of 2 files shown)


robjtede:
released with v0.6.2

PR #249, opened 2022-05-24
Return `&str` from `BasicAuth::user_id()` and `BasicAuth::password()`

### LINKED ISSUE #238: actix-web-httpauth: Basic::{user_id,password} return reference to Cow
The [`Basic`](https://docs.rs/actix-web-httpauth/latest/actix_web_httpauth/headers/authorization/struct.Basic.html#) struct has two getter methods, `user_id` and `password`, both of which return a `&Cow<'static, str>` (one wrapped in an option). This is a big ergonomics papercut. Partly the getters become tied to the underlying implementation of the struct, removing most of the point of having a private field with a public getter, and partly the extra type noise makes the returned value useful in less cases. It would be much better to return a `&str` (and `Option<&str>`) directly instead.

mcronce:
## PR Type
Feature/improvement

## PR Checklist

## Overview
Makes the two mentioned functions more easily usable in calling code

I would probably consider this a breaking change - the return types of public functions have changed.  It _might_ be the case that any code that compiled/worked previously will still compile, but I'm not super confident about that.

Fixes #238


JohnTitor:
Does it make sense to apply the same change to `Bearer::token`?

https://github.com/actix/actix-extras/blob/d11a2723841d1eb376c1fbfbda8f432ad2d7d9c6/actix-web-httpauth/src/headers/authorization/scheme/bearer.rs#L37-L40

PR #253, opened 2022-06-10
Added function to create an empty session for testing

### LINKED ISSUE #252: Feature request: implement Default or add a constructor to Session for testing
In my project, I created a trait and implemented it for the Session struct that added some basic functionality customized to my use case (in this case, validating a user id and expiration time against the SessionStore, for authentication purposes).

I wanted to package this trait as a library to use in multiple microservices, and build unit tests for it. But I am getting stuck in that I cannot see a way to create an instance of Session for my tests. There is no constructor, the `Session{...}` syntax does not work due to private fields, and I can't seem to get a way to create one via an HttpRequest (which also does not appear to have a constructor) within a unit test. 

I'm going to propose a PR to do this shortly, but I wanted to add an issue explaining what I was doing.

kingledion:
## PR Type

Feature

## PR Checklist

- [ ] Tests for the changes have been added / updated. (Not applicable, no tests existed)

## Overview

Added a `Session::empty()` function which instantiates an empty session to simplify unit testing. 

None expected

Closes #252


kingledion:
I dunno what to say about this. I can't replicate these build failures locally, and they also appear to be in actix-cors which I didn't change.

I tested a variety of ways that edition 2021/2018 and rust 1.54/1.60 may have interacted to cause a dependency issues, and I can't find anything. 

The download of hashbrown failure may be an availability issue? Perhaps the solution is to re-run pipeline? I don't see that I can trigger it again...

JohnTitor:
Nah, hashbrown started to use edition 2021 and it means we have to bump up our MSRV to 1.56. I'll handle it later so you don't have to worry about it.

robjtede:
#233 is a big change and will likely conflict with this MR but I'd like to get it merged first (very soon).

robjtede:
I can believe that a function like this is still useful after the overhaul but I'd like to see a test added demonstrating one or two use cases.

PR #265, opened 2022-07-21
fix: bad interaction between session state changes and renewal

### LINKED ISSUE #244: Bad interaction between session state changes and renewal
We suffer from the same bug reported against `poem`'s `Session` implementation - see https://github.com/poem-web/poem/issues/196

Hulxv:
## PR Type

Bug Fix

## PR Checklist

## Overview

Closes  #244

### DIFF
--- a/actix-session/CHANGES.md
+++ b/actix-session/CHANGES.md
@@ -1,7 +1,9 @@
 # Changes
 
 ## Unreleased - 2021-xx-xx
+- Fix bad interaction between session state changes and renewal. [#265]
 
+[#265]: https://github.com/actix/actix-extras/pull/265
 
 ## 0.7.0 - 2022-07-09
 - Added `TtlExtensionPolicy` enum to support different strategies for extending the TTL attached to the session state. `TtlExtensionPolicy::OnEveryRequest` now allows for long-lived sessions that do not expire if the user remains active. [#233]
--- a/actix-session/src/session.rs
+++ b/actix-session/src/session.rs
@@ -128,7 +128,9 @@ impl Session {
         let mut inner = self.0.borrow_mut();
 
         if inner.status != SessionStatus::Purged {
-            inner.status = SessionStatus::Changed;
+            if inner.status != SessionStatus::Renewed {
+                inner.status = SessionStatus::Changed;
+            }
 
             let key = key.into();
             let val = serde_json::to_string(&value)
@@ -155,7 +157,9 @@ impl Session {
         let mut inner = self.0.borrow_mut();
 
         if inner.status != SessionStatus::Purged {
-            inner.status = SessionStatus::Changed;
+            if inner.status != SessionStatus::Renewed {
+                inner.status = SessionStatus::Changed;
+            }
             return inner.state.remove(key);
         }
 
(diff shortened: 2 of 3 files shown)


robjtede:
For the sake of readers like me who haven't really looked into the problem, is there a comment that can be left on these conditionals explaining the behavior? Not obvious at first glance (to me) what this solves.

Hulxv:
That's a simple explanation for the issue from [poem-web/poem#196](https://github.com/poem-web/poem/issues/196) and that issue happens in `actix-session` too.

> ## Expected Behavior
> 
> When using Session, calling `session.renew()` always renews the session, indepentent of it's location in the handler (unless there is a `session.purge()`).
> 
> ## Actual Behavior
> When calling `session.set("x", "y")` after `session.renew()`, the session does not get renewed.
> 
> From my understanding of the code, this is caused by this line in `session.set()` (and `session.remove()`):
> https://github.com/poem-web/poem/blob/ad20fe7ebaf9660a295d1af194ee82b6d2d80037/poem/src/session/session.rs#L85
> 
> This unconditionally overwrites the status and deletes the information that a session renew was requested. What I think should happen instead is that it only overwrites the status if it is currently set to `Unchanged`.

robjtede:
Ok, got it. So, if I understand correctly, this won't break the other way because a renewed session will reassign the cookie with the changed state as well.

Hulxv:
> is should make the state more like setting independent boolean flags than an enum? 

Nope, Enums is working well and I think I resolved the issue that is happening.

@LukeMathWalker Can you give us your opinion?

LukeMathWalker:
I think @robjtede's intuition is correct - renewing the session id and changes the session state are independent and should be tracked separately.

Hulxv:
So is there anything I should do more or is everything just fine?

robjtede:
My concern is that this might break it the other way; storage implementations handle independently a change in session content and session lifespan.

It would not affect cookie storage but in Redis storage is would be the different... eg:
```
CHANGED           => SET key new_contents
RENEWED           => TTL key ttl
CHANGED + RENEWED => SETX key new_contents ttl
```
^ untested but hopefully you see the point

robjtede:
Here's what I want to know:
1. Does this fix the issue #244 without introducing new bugs?
2. Is this the ideal set up of session state and TTL metadata?

Hulxv:
> 1. Does this fix the issue #244 without introducing new bugs?

I think yes because that is the way how `poem`  developers solved this issue and there's no one reported another issue about that more and they never change [session.rs](https://github.com/poem-web/poem/blob/master/poem/src/session/session.rs) after [20532aa](https://github.com/poem-web/poem/commit/20532aa06bcbb4bc122d67d3a4fc0b5fcf68eef4).

> 2. Is this the ideal representation of session state and TTL metadata?


Honestly, I'm don't familiar with `Redis`, So, I don't understand what you mean here
```
CHANGED           => SET key new_contents KEEPTTL
RENEWED           => SETEX key ttl
CHANGED + RENEWED => SET key new_contents EX ttl
```
and I don't know what the ideal representation.

LukeMathWalker:
This should not introduce new bugs, as far as my understanding goes. 

The middleware instructs the backend to perform a delete on the data associated with the old session key and a save of the current session state against the new key (see https://github.com/actix/actix-extras/blob/cfd16c54781a154f937bbbddf605bb0ce7a2e564/actix-session/src/middleware.rs#L275), so this should behave correctly regardless of the session storage implementation details.

robjtede (approved):

PR #270, opened 2022-07-31
adopt actix-settings crate

robjtede:
## PR Type

Crate Adoption

## PR Checklist

## Overview
Adopt https://github.com/jjpe/actix-settings into the @actix org as discussed here: https://github.com/jjpe/actix-settings/issues/1.

Also updates it to v4 ecosystem.

cc @jjpe

### DIFF
--- a/actix-settings/src/lib.rs
+++ b/actix-settings/src/lib.rs
@@ -1,0 +1,695 @@
+/// A library to process Server.toml files
+use std::{
+    env, fmt,
+    fs::File,
+    io::{Read as _, Write as _},
+    path::Path,
+    time::Duration,
+};
+
+use actix_http::{Request, Response};
+use actix_service::IntoServiceFactory;
+use actix_web::{
+    body::MessageBody,
+    dev::{AppConfig, ServiceFactory},
+    http::KeepAlive as ActixKeepAlive,
+    Error as WebError, HttpServer,
+};
+use serde::Deserialize;
+
+#[macro_use]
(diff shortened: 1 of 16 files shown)


actix-settings/src/lib.rs
```diff
@@ -0,0 +1,705 @@
...
+        }
+    }
+}
+
+pub trait ApplySettings {
+    /// Apply a [`BasicSettings`] value to `self`.
+    ///
+    /// [`BasicSettings`]: ./struct.BasicSettings.html
+    #[must_use]
+    fn apply_settings<A>(self, settings: &BasicSettings<A>) -> Self
```
jjpe, line 129: Perhaps the return type should be altered to return some kind of `Result<Self, ...>`.
That way, the implementing method in the `impl ApplySettings for HttpServer` below can properly propagate its error rather than simply `unwrap()` ing it. That error can only happen if binding 1 or more addresses fails, which for valid addresses generally isn't an issue.
But technically it's still a landmine I guess.

jjpe (approved):

actix-settings/src/lib.rs
robjtede: Yeah. Probably better to leave it for now and pencil it for a subsequent release.

jjpe: Works for me. By that time, the error should probably be a new `AtError` variant.

jjpe (approved):

PR #271, opened 2022-07-31
document settings crate

robjtede:
## PR Type

Docs

## PR Checklist

- [x] ~~A changelog entry has been made for the appropriate packages.~~

## Overview
Docuemnt crate in preparation for publishing new version.

cc @jjpe lmk if anything is inaccurate.

### DIFF
--- a/actix-settings/examples/actix.rs
+++ b/actix-settings/examples/actix.rs
@@ -62,7 +62,7 @@ async fn main() -> std::io::Result<()> {
     .await
 }
 
-/// Initialize the logging infrastructure
+/// Initialize the logging infrastructure.
 fn init_logger(settings: &Settings) {
     if !settings.actix.enable_log {
         return;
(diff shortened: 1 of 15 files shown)


robjtede:
@jjpe can you shed some light on the naming of `AtError` ?

robjtede:
Ahh it'll be an `actix-toml` abbreviation (as seen in your repo's readme). I assume it was a working title?

jjpe:
> I assume that was the working title.

Pretty much yeah. So the error should probably be renamed to either `Error` (letting the module name do the scoping) or to something like `SettingsResult` or `AsResult`. Personally I'd favor the first, as it's a common idiom, and the last one could easily get confused for a trait. The main issue with `SettingsResult` is the wordiness I think.

robjtede:
My preference is also to just use `Error`. And likely make the `Result` convenience private. But again, I'll defer other API changes to a subsequent release.

PR #280, opened 2022-08-26
Limitation: display and handle client error

GreeFine:
## PR Type

Feature

## PR Checklist

- [x] ~~Tests for the changes have been added / updated.~~
- [x] ~~Documentation comments have been added / updated.~~

## Overview
Currently: if any error occur during the count process the response will always be: "Rate limit exceed"
New behavior: errors from the count function are matched and respond with **500 Internal server error** if it's an unexpected error

I wanted to add clarity on what happened during the count process, because when trying out this crate I didn't understand why things worked/didn't work and all I was receiving was "Rate limit exceed" which wasn't helpful in understanding what was the issue

I am not sure that responding with "Internal server error" is not too severe, but at it prevent the limitation from working properly for me I prefer this than the silent error.


GreeFine:
I am not sure what to do for " Tests for the changes have been added / updated.".
I don't know how I could test what I added with the current setup. And there is no existing tests on that part of the crate

robjtede (commented):
good fix, thanks 👍🏻

Issue #239, opened 2022-04-02
actix-limitation: Needs a way to set a custom key


jacob-pro:
There doesn't appear to be a way to set a custom key for actix-limitation.

For example I would like to build a rate limit key based on the user's remote IP address for the purpose of protecting a login route.

It would also be good to remove the required dependency on actix-session because depending on how you are generating the key you might not need to use it.

jacob-pro:
Update:
I have written my own rate limiter implementation so I'm no longer needing this to be implemented here
https://github.com/jacob-pro/actix-extensible-rate-limit

PR #281, opened 2022-08-26
Limitation: custom key from closure

GreeFine:
## PR Type

Feature

## PR Checklist

## Overview
Currently the key used to identify a user is either done with actix-session or with a cookie. 
Made the Limiter take a closure do define any custom key, for example using the IP as a key 

I tried to keep backward compatibility in mind but deprecated the old functions to promote the new feature 

closes https://github.com/actix/actix-extras/issues/239


GreeFine:
@robjtede I have fixed the issue with the cargo toml, update the changelog and update the doc

actix-limitation/src/lib.rs
```diff
@@ -70,16 +77,27 @@ pub const DEFAULT_PERIOD_SECS: u64 = 3600;
...
+pub type GetKeyFn = Box<dyn GetKeyFnT + Send + Sync>;
+impl fmt::Debug for GetKeyFn {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "GetKeyFn")
+    }
+}
+
 /// Rate limiter.
-#[derive(Debug, Clone)]
+#[derive(Debug)]
```
robjtede, line 95: This is the only reason I've seen to mark the PR breaking, if GetKeyFn could be Rc instead of Box it could be semver-minor

GreeFine: Oh ! I forgot about this, I wrapped the Closure field in an Arc, 
```rs
    get_key_fn: Arc<GetKeyFn>,
```
Is this what you wanted me to do ?

GreeFine: :sweat: Rc doesn't work because it's not Send
https://github.com/actix/actix-extras/runs/8071465240?check_suite_focus=true#step:6:45

I reverted back to the original Arc instead.

actix-limitation/src/builder.rs
```diff
@@ -1,40 +1,60 @@
...
     /// Set upper limit.
-    pub fn limit(&mut self, limit: usize) -> &mut Self {
+    pub fn limit(mut self, limit: usize) -> Self {
         self.limit = limit;
         self
     }
 
     /// Set limit window/period.
-    pub fn period(&mut self, period: Duration) -> &mut Self {
+    pub fn period(mut self, period: Duration) -> Self {
```
robjtede, line 30: This signature change is also breaking.

To be clear, i'm fine with breaks but just pointing it out given what you said. It's also a bit more common within the Actix ecosystem to use the `-> &mut Self` builder pattern (see `HttpResponseBuilder`).

actix-limitation/src/lib.rs
```diff
@@ -70,16 +78,27 @@ pub const DEFAULT_PERIOD_SECS: u64 = 3600;
...
 
 /// Default session key.
+#[cfg(feature = "session")]
 pub const DEFAULT_SESSION_KEY: &str = "rate-api-id";
 
+/// Helper trait to impl Debug on GetKeyFn type
+pub trait GetKeyFnT: Fn(&ServiceRequest) -> Option<String> + Sync {}
+impl<T> GetKeyFnT for T where T: Fn(&ServiceRequest) -> Option<String> + Sync {}
+/// Get key resolver function type
+pub type GetKeyFn = Box<dyn GetKeyFnT + Send + Sync>;
```
robjtede, line 88: I think more so I meant using Rc/Arc here.

actix-limitation/src/builder.rs
GreeFine: I reverted back to `&mut Self` 👍 
(now that it's wraped in Arc the closure is clonable and I can use an &mut self)

If you have the time would you mind explaining me why is Actix using `&mut Self` on builder pattern instead of `Self` ?

robjtede: It allows more options for configuring, both chaining and stand-alone mutation.

```
let limiter = Limiter::builder("redis://127.0.0.1").limt(500).build().unwrap();

// and

let mut limiter = Limiter::builder("redis://127.0.0.1");
limiter.limt(500);
let limiter = limiter.build().unwrap();
```

The latter being more difficult to integrate with conditional settings when using consuming builders.

robjtede (approved):

PR #285, opened 2022-09-12
Add note for accessing session state in stream.

mohe2015:
## PR Type

Small documentation improvement.

## PR Checklist

## Overview

I found out that you can't even read the session state in a stream of a streaming response. This may be unintentional behaviour as I think reading could be possible just not writing as the headers have already been sent. In any case I think it may be helpful to document this.

### DIFF
--- a/actix-session/src/lib.rs
+++ b/actix-session/src/lib.rs
@@ -65,7 +65,7 @@
 //! ```
 //!
 //! The session state can be accessed and modified by your request handlers using the [`Session`]
-//! extractor.
+//! extractor. Note that this doesn't work in the stream of a streaming response.
 //!
 //! ```no_run
 //! use actix_web::Error;


robjtede:
I don't really understand what you're expecting to work here. Can you give a psuedo-code example of what you want to write?

mohe2015:
> I don't really understand what you're expecting to work here. Can you give a psuedo-code example of what you want to write?

To be honest my explanation was a bit short. I wrote a reproducer in https://github.com/mohe2015/actix-extras-285/blob/main/src/main.rs

If I didn't mess anything up in a perfect world I would expect the set-broken and get-broken to either not compile or to panic. To my understanding they silently don't contain the data of the session. It is clear though that setting the data is not supposed to work as the headers have already been sent (potentially).

robjtede (approved):

robjtede:
Thanks for showing how you're wanting to use it, the added note makes more sense now.

mohe2015:
A question I have is whether I should open an issue to improve this behaviour (especially on the reading side)?

PR #287, opened 2022-09-18
Add block_on_origin_mismatch option to middleware

### LINKED ISSUE #286: actix_cors blocks requests that should be allowed
## Expected Behavior

If I have app.example.com as a whitelisted Origin but both the backend + frontend is loaded from app2.example.com, request should be allowed.

## Current Behavior

If I have app.example.com as a whitelisted Origin but both the backend + frontend is loaded from app2.example.com, requests are blocked, even though Host + Origin is the same. Browsers allow these requests to happen (no OPTIONS request sent, as Origin + Host is the same), but actix_cors decides to forcefully block the request.

Relevant code from actix_cors:

https://github.com/actix/actix-extras/blob/1561bda82268931edb070501413a46337ed71290/actix-cors/src/inner.rs#L73-L99

## Possible Solution

actix_cors should not forcefully cancel/error on requests where Origin and Host is the same header. In fact, if there is no OPTIONS request, actix_cors should not care about the actual request at all, besides setting the right headers on the Response. As it stands today, actix_cors is trying to be overly protective of responses while it doesn't add anything regarding security.

CapableWeb:
## PR Type

Bug Fix

## PR Checklist

- [X] Tests for the changes have been added / updated.
- [X] Documentation comments have been added / updated.
- [X] A changelog entry has been made for the appropriate packages.
- [X] Format code with the nightly rustfmt (`cargo +nightly fmt`).

## Overview

Retains current behavior, adds a new option to get the behaviour described in the issue, where it will be up to browsers to handle if the request should be allowed or not, allowing other tools like cURL (who don't care about Origin) to continue to work regardless of CORS headers. 

Closes #286

### DIFF
--- a/actix-cors/CHANGES.md
+++ b/actix-cors/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Unreleased - 2022-xx-xx
 - Minimum supported Rust version (MSRV) is now 1.59 due to transitive `time` dependency.
+- Add `block_on_origin_mismatch` option to actix-cors [#286]
 
 
 ## 0.6.2 - 2022-08-07
--- a/actix-cors/CHANGES.md
+++ b/actix-cors/CHANGES.md
@@ -2,7 +2,9 @@
 
 ## Unreleased - 2022-xx-xx
 - Minimum supported Rust version (MSRV) is now 1.59 due to transitive `time` dependency.
-- Add `block_on_origin_mismatch` option to actix-cors [#286]
+- Add `Cors::block_on_origin_mismatch()` option for controlling if requests are pre-emptively rejected. [#286]
+
+[#286]: https://github.com/actix/actix-extras/pull/286
 
 
 ## 0.6.2 - 2022-08-07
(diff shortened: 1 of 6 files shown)


robjtede:
Changed some stuff:
- Tweaked wording on docs
- Used early return style on the boolean check
- Showed usage of the new option in the example
- Added some more asserts to the tests

On that last point, the tests now fail. The reason is that this change opens up some subtle assuptions that actix-cors has been able to make until now. You can see from using new example the following response:

```console
$ http :8080/ Origin:wrong.com
HTTP/1.1 200 OK
access-control-allow-origin: wrong.com
access-control-expose-headers: content-disposition
content-length: 26
content-type: text/plain; charset=utf-8
date: Mon, 19 Sep 2022 21:34:49 GMT
vary: Origin, Access-Control-Request-Method, Access-Control-Request-Headers

Hello, cross-origin world!
```

In particular, `access-control-allow-origin: wrong.com` shows up despite not being an allowed origin so this is currently incorrect behavior. This behavior changing so that origin validation (and maybe other validation) is done leading to conditionally adding CORS response headers. The new test assert should serve as a validator for further changes to this PR.

robjtede (approved):

PR #292, opened 2022-10-04
[LIMITATION] Feature: allow multiple Limiters with different configurations

GreeFine:
## PR Type

Feature

## PR Checklist

## Overview

Currently I can declare and use only one Limiter who can hold one configuration
This PR adds the possibility to create a multitude of Limiter so we can have different configurations

## Notes
I added a method to build the Limiter with an already existing Client (`builder_with_redis_client`) to prevent re-creating a new client for each new Limiter
I don't actually know if this was a good idea and is making any performance improvements. And I didn't have the time to test it

### DIFF
--- a/actix-limitation/src/middleware.rs
+++ b/actix-limitation/src/middleware.rs
@@ -1,4 +1,4 @@
-use std::{future::Future, pin::Pin, rc::Rc};
+use std::{collections::HashMap, future::Future, pin::Pin, rc::Rc};
 
 use actix_utils::future::{ok, Ready};
 use actix_web::{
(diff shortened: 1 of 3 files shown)


actix-limitation/src/middleware.rs
```diff
@@ -11,9 +11,23 @@ use actix_web::{
...
 /// Rate limit middleware.
+///
+/// Use the `scope` variable to define multiple limiter
 #[derive(Debug, Default)]
 #[non_exhaustive]
-pub struct RateLimiter;
+pub struct RateLimiter {
+    /// Used to define multiple limiter, with different configurations
+    ///
+    /// WARNING: When used (not None) the middleware will expect a `HashMap<Limiter>` in the actix-web `app_data`
```
JohnTitor: This doesn't match with the implementation.

PR #296, opened 2022-10-26
Feature: Add IdentityError to actix-identity crate.

### LINKED ISSUE #272: Actix Identity anyhow error type is incompatible with actix_web::Error
The error type used by actix_identity is incompatible with acitx_web::Error, creating some annoying usability problems inside actix_web code.

## Expected Behavior
I can just perform
```rust
pub async fn foo(identity: Identity) -> Result<HttpResponse, actix_web::Error> {
    identity.id()?;
```

## Current Behavior
Above code will fail as id() will simply return anyhow::Error, which isn't convertible to actix_web::Error and also an indirect re-export of anyhow, which isn't that good of a choice for libraries to use (over thiserror for example).

## Possible Solution
Export some error type (not anyhow at best), that has an impl for actix_web::Error or can be easily converted. This was possible in earlier actix_identity.

1. See code above

## Context

I've got to manually convert each invocation of identity.id() or create a super type that wraps all these errors, while still maintaining `impl ResponseError`.

## Your Environment

* Rust Version (I.e, output of `rustc -V`): 1.62.1
* Actix-* crate(s) Version:
```toml
actix = "0.13"
actix-web = "4"
actix-rt = "2"
actix-files = "0.6"
actix-identity = "0.5"
actix-session = { features = ["cookie-session"], version = "0.7" }
actix-threadpool = "0.3"
```

esmevane:
## PR Type

Feature

## PR Checklist

## Overview

In order to let crates in the actix web ecosystem interact correctly with `actix_web::Error`, this commit introduces its own error type, replacing the previous usage of `anyhow::Error`.

This will break any consumer code that depends on the presence of `anyhow::Error`, but in exchange it will allow more users of `actix-identity` to leverage its public API inside responders with less work.

Closes #272


GreeFine (approved):
Looks good 👍 

There is only one thing that I would add is the impl for `actix_web::error::response_error::ResponseError` as it was asked in the issue

actix-identity/src/identity_error.rs
```diff
@@ -0,0 +1,62 @@
+use actix_session::{SessionGetError, SessionInsertError};
+use actix_web::cookie::time::error::ComponentRange;
+
+/// Possible errors which can emerge from storing and retrieving identities.
+#[derive(Debug)]
```
GreeFine, line 5: You could add the `#[non_exhaustive]` tag, so we can add new errors latter without breaking libraries using it

esmevane: Sweet - I've split the errors, but put `#[non_exhaustive]` wherever I left an enum

actix-identity/src/identity_errors.rs
```diff
@@ -0,0 +1,69 @@
...
+    fn from(error: SessionInsertError) -> Self {
+        Self(error)
+    }
+}
+
+/// This error describes all of the potential failures which can happen
+/// while retrieving an identity.
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum IdentityError {
```
LukeMathWalker, line 32: I'd suggest naming this `GetIdentityError`.

actix-identity/src/identity.rs
```diff
@@ -126,9 +124,12 @@ impl Identity {
     ///     }
     /// }
     /// ```
-    pub fn id(&self) -> Result<String, anyhow::Error> {
+    pub fn id(&self) -> Result<String, IdentityError> {
         self.0.session.get(ID_KEY)?.ok_or_else(|| {
-            anyhow!("Bug: the identity information attached to the current session has disappeared")
+            IdentityError::MissingIdentityError(
+                "Bug: the identity information attached to the current session has disappeared"
```
LukeMathWalker, line 130: I don't think this should be treated in the same way of an error in `get_identity`.
When `id` is called, we have already successfully extracted the identity from the current session once. I'd suggest using a different error type here.

actix-identity/src/identity_errors.rs
```diff
@@ -0,0 +1,69 @@
...
+    /// This occurs whenever no identity is found in a session.
+    MissingIdentityError(String),
```
LukeMathWalker, line 34: This doesn't need to hold a `String` if we use a different error type for the `id` method.
We can simply provide the correct error message in the `Display` implementation.

actix-identity/src/identity_errors.rs
```diff
@@ -0,0 +1,69 @@
...
+#[derive(Debug)]
+pub struct LoginError(SessionInsertError);
+
+impl std::fmt::Display for LoginError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+impl std::error::Error for LoginError {}
```
LukeMathWalker, line 14: We must implement the `source` method and return the wrapped error.

actix-identity/src/identity_errors.rs
```diff
@@ -0,0 +1,69 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::MissingIdentityError(message) => write!(f, "{}", message),
+            Self::SessionExpiryError(source) => write!(f, "{}", source),
+            Self::SessionGetError(source) => write!(f, "{}", source),
+        }
+    }
+}
+
+impl std::error::Error for IdentityError {}
```
LukeMathWalker, line 51: We must implement the `source` method and return the wrapped errors.

actix-identity/src/identity.rs
esmevane: Is it worth decoupling the get operation and the bug handling instance? It seems like our choices are either split the interface, allowing for two error return types, or have an enum describing both potential error outcomes.

(Also totally possible I'm misunderstanding something)

actix-identity/src/identity_errors.rs
esmevane: Oh! I didn't realize. Easy enough to do  :)

actix-identity/src/identity_errors.rs
```diff
@@ -0,0 +1,91 @@
...
+pub enum GetIdentityError {
+    /// This is an error which shouldn't occur, and indicates some kind of bug.
+    LostIdentityError,
+    /// This occurs whenever no identity is found in a session.
+    MissingIdentityError,
+
+    /// This occurs whenever something goes wrong accessing a session store.
+    SessionGetError(SessionGetError),
+    /// This occurs whenever any kind of expiration of a session has taken place.
+    SessionExpiryError(ComponentRange),
```
LukeMathWalker, line 45: I'd prefer not to expose `ComponentRange` here - it'd be better to introduce a `SessionExpiryError` struct that wraps `ComponentRange` and doesn't expose it.

actix-identity/src/identity_errors.rs
```diff
@@ -0,0 +1,91 @@
...
+
+/// This error describes all of the potential failures which can happen
+/// while retrieving an identity.
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum GetIdentityError {
+    /// This is an error which shouldn't occur, and indicates some kind of bug.
+    LostIdentityError,
+    /// This occurs whenever no identity is found in a session.
+    MissingIdentityError,
```
LukeMathWalker, line 40: For forward-compatibility - let's make this variant non-unit (i.e. they wrap a struct) and that struct is a unit struct marked with #[non_exhaustive]. It will allow us to add fields in the future in a non-breaking fashion.

actix-identity/src/identity_errors.rs
esmevane: Cool - I _think_ I've done enough to keep `ComponentRange` behind that struct, let me know if I missed anything there?

actix-identity/src/identity_errors.rs
esmevane: Ok! I am constructing and coercing it in the `get_identity` implementation. Is that the right way to hand it outward? `non_exhaustive` seems like it only cares if something outside the crate tries to construct it, but I want to be sure it's ok to do it in a neighbor module.

actix-identity/src/identity_errors.rs
```diff
@@ -0,0 +1,130 @@
...
+    }
+}
+
+impl From<SessionInsertError> for LoginError {
+    fn from(error: SessionInsertError) -> Self {
+        Self(error)
+    }
+}
+
+/// A wrapper for [ComponentRange] which adds expiration context.
```
LukeMathWalker, line 32:
```suggestion
/// An error encountered when working with a session that has expired.
```

actix-identity/src/identity_errors.rs
```diff
@@ -0,0 +1,130 @@
...
+    }
+}
+
+impl std::error::Error for SessionExpiryError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        Some(&self.0)
+    }
+}
+
+/// A marker left for any identity error nuance we may want to communicate.
```
LukeMathWalker, line 48:
```suggestion
/// The identity information in the current session has disappeared after having been successfully validated.
```

actix-identity/src/identity_errors.rs
```diff
@@ -0,0 +1,130 @@
...
+#[derive(Debug)]
+#[non_exhaustive]
+pub struct MissingIdentityError;
+
+impl std::fmt::Display for MissingIdentityError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self)
```
LukeMathWalker, line 55:
```suggestion
        write!(f, "The identity information in the current session has disappeared after having been successfully validated. This is likely to be a bug.")
```

actix-identity/src/identity_errors.rs
```diff
@@ -0,0 +1,130 @@
...
+    /// This occurs whenever any kind of expiration of a session has taken place.
+    SessionExpiryError(SessionExpiryError),
+}
+
+impl std::fmt::Display for GetIdentityError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::LostIdentityError => write!(
+                f,
+                "Bug: the identity information attached to the current session has disappeared"
```
LukeMathWalker, line 86:
```suggestion
                "The identity information in the current session has disappeared after having been successfully validated. This is likely to be a bug."
```

actix-identity/src/identity_errors.rs
```diff
@@ -0,0 +1,130 @@
...
+            Self::SessionExpiryError(source) => write!(f, "{}", source),
+            Self::SessionGetError(source) => write!(f, "{}", source),
+        }
+    }
+}
+
+impl std::error::Error for GetIdentityError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        match self {
+            Self::LostIdentityError | Self::MissingIdentityError(_) => None,
```
LukeMathWalker, line 101:
```suggestion
            Self::LostIdentityError => None,
            Self::MissingIdentityError(source) => Some(source),
```

actix-identity/src/identity_errors.rs
LukeMathWalker: Yes, the way you did is perfect 👌🏻 
We just need to do the same for `LostIdentityError` and then we are done.

actix-identity/src/identity_errors.rs
LukeMathWalker: It's looking good!

actix-identity/src/identity_errors.rs
LukeMathWalker: Do you have time to drive this to the finish line @esmevane? Or should we merge it as is and then add the remaining small changes to `LostIdentityError`?

esmevane: @LukeMathWalker Ah, mea culpa! I think the holidays got me. Stand by for this :)

LukeMathWalker (approved):

LukeMathWalker:
Correct. A maintainer with commit access (most likely @robjtede) must now decide to move forward with this.

esmevane:
Perfect! Thanks again for the 👀 & tips

robjtede (approved):

robjtede:
Apologies for the delay and thanks the contribution!

PR #300, opened 2022-11-08
Add secure field to removal cookie

### LINKED ISSUE #299: Secure field never set in removal cookie.
Secure field is never being set on removal cookie in actix-session. When using the __Host- or __Secure- cookie naming scheme, clearing the cookie will not work since it requires the secure field.

https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-cookie-prefixes-00#section-3

## Expected Behavior

When sending the removal cookie with empty value it should clear the cookie in the browser.

## Current Behavior

Since the secure field is not set, chrome does not overwrite the cookie.

## Possible Solution

Setting the secure field in the cookie created in  `fn delete_session_cookie` in `actix-session/src/middleware` like we set the other attributes.

1. Use a cookie with a __Host- or __Secure- prefix.
2. Try logging out using `Identity::logout`.
3. Chrome will show a warning telling that setting this cookie was blocked because it breakes the rules of the prefixes.
4.The cookie will not be cleared.

## Your Environment

- Chrome Version 107.0.5304.87 (Official Build) (64-bit)

* Rust Version (I.e, output of `rustc -V`): rustc 1.65.0 (897e37553 2022-11-02)
* Actix-* crate(s) Version:
actix-identity = "0.5"
actix-session = { version = "0.7", features = ["redis-rs-tls-session"] }
actix-web = "4"

evenorog:
## PR Type

Bug Fix

## PR Checklist

## Overview

Adds the secure field to removal cookie required for cookies with __Host- and __Secure- prefixes.

Closes #299

### DIFF
--- a/actix-session/CHANGES.md
+++ b/actix-session/CHANGES.md
@@ -1,7 +1,9 @@
 # Changes
 
 ## Unreleased - 2021-xx-xx
-- Set secure attribute when adding a session removal cookie.
+- Set secure attribute when adding a session removal cookie. [#300]
+
+[#300]: https://github.com/actix/actix-extras/pull/300
 
 ## 0.7.2 - 2022-09-11
 - Set SameSite attribute when adding a session removal cookie. [#284]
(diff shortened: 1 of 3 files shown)


robjtede (commented):
thanks 👍🏻

robjtede (approved):

Issue #306, opened 2022-12-04
Ability to extract SessionKey


robjtede:
### Discussed in https://github.com/actix/actix-extras/discussions/305

<div type='discussions-op-text'>

<sup>Originally posted by **miketwenty1** December  4, 2022</sup>
**Goal:** Prevent 1 user from having multiple active sessions.

I'm using actix-web redis session middleware.
```
.wrap(SessionMiddleware::new(
    RedisActorSessionStore::new(redis_conn.clone()),
    secret_key.clone(),
))
```

When a user does a login, I'll go ahead and insert some data into their session.
```
session.insert("userid", id).expect("id inserted");
```

Seemingly these key/values via the `session.insert` are inner key/values to the actual managed middleware session keys that are made automagically.

**First question:** How can I extract this middleware session id key that redis uses for the inserts/sessions?

My thought is I would go ahead and insert a the userid as a redis key with the value being the middleware redis session id so I could easily lookup and void any sessions that the user may have if logging in again. (seems hacky idk).

**Second question:** Is there a much easier/cleaner way of doing this while still getting the nice managed sessions from the redis middleware?</div>

robjtede:
@LukeMathWalker any initial thoughts on this?

It seems like a method like the following would satisfy this use case, punting the actual purge logic to the application.

```rust
fn key(&self) -> SessionKey
```

LukeMathWalker:
That would work for the Redis case, yes. I'd suggest returning a `secrecy::Secret<SessionKey>` though.

miketwenty1:
@LukeMathWalker do you envision `secrecy::Secret<SessionKey>` functionality only being added for redis actor?
If so, would this only be changes to redis_actor.rs?

LukeMathWalker:
No, it should be added to the `Session` struct, as @robjtede suggested.
I was just mentioning that achieving the same objective (limiting the number of concurrent sessions) with other backends (i.e. cookies) won't work in this way since the `SessionKey` changes based on the data attached to the session.

miketwenty1:
Is this something you would be interested in working on with me? 
I'm sort of a beginner but I'm interested in getting this feature and contributing with some guidance if needed.

I imagine this would need actually need to take place in the inner struct?
```rust
#[derive(Default)]
struct SessionInner {
    state: HashMap<String, String>,
    status: SessionStatus,
    session_key: secrecy::Secret<SessionKey>
}
```
?

LukeMathWalker:
[...]

PR #307, opened 2022-12-14
Feature/session key

miketwenty1:
## PR Type

INSERT_PR_TYPE:
Code Style: Cargo.toml
Feature: Session Key Extraction ability.

## PR Checklist

## Overview

The goal is add this feature to be utilized by the redis session middleware without breaking any other session middleware. 

Notes from discussion:
https://github.com/actix/actix-extras/issues/306

Dev Notes:
This is not a working PR, changes will come after feedback.

### DIFF
--- a/actix-session/src/session.rs
+++ b/actix-session/src/session.rs
@@ -17,6 +17,8 @@ use anyhow::Context;
 use derive_more::{Display, From};
 use serde::{de::DeserializeOwned, Serialize};
 
+use crate::storage::SessionKey;
+
 /// The primary interface to access and modify session state.
 ///
 /// [`Session`] is an [extractor](#impl-FromRequest)—you can specify it as an input type for your
@@ -77,6 +79,7 @@ impl Default for SessionStatus {
 struct SessionInner {
     state: HashMap<String, String>,
     status: SessionStatus,
+    session_key: SessionKey,
 }
 
 impl Session {
@@ -101,7 +104,13 @@ impl Session {
             Ok(None)
         }
     }
-
+    /// Get a the session key itself from the overall session.
+    ///
+    /// Needs to be implemented
+    pub fn get_session_key(&self) -> secrecy::Secret<SessionKey> {
+        let key = self.0.borrow().session_key.clone();
+        secrecy::Secret::new(key)
+    }
     /// Get all raw key-value data from the session.
     ///
     /// Note that values are JSON encoded.
--- a/actix-session/src/session.rs
+++ b/actix-session/src/session.rs
@@ -106,9 +106,9 @@ impl Session {
     }
     /// Get a the session key itself from the overall session.
     ///
-    /// Needs to be implemented
+    /// Retrieve the overall session key
     pub fn get_session_key(&self) -> secrecy::Secret<SessionKey> {
-        let key = self.0.borrow().session_key.clone();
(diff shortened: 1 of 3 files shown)


actix-session/src/session.rs
```diff
@@ -77,6 +79,7 @@ impl Default for SessionStatus {
 struct SessionInner {
     state: HashMap<String, String>,
     status: SessionStatus,
+    session_key: SessionKey,
```
LukeMathWalker, line 82: This must be an `Option`, since the key will only exist if the user has an active session.

actix-session/src/storage/session_key.rs
```diff
@@ -17,8 +17,8 @@ use derive_more::{Display, From};
 /// let session_key: Result<SessionKey, _> = key.try_into();
 /// assert!(session_key.is_err());
 /// ```
-#[derive(Debug, PartialEq, Eq)]
-pub struct SessionKey(String);
+#[derive(Debug, PartialEq, Eq, Default, Clone)]
+pub struct SessionKey(pub String);
```
LukeMathWalker, line 21: You don't want to expose the inner field here. 
We can change `SessionKey` to be `SessionKey(secrecy::Secret<String>)` and expose `into_inner`/`inner` methods to access the inner value.

actix-session/src/session.rs
LukeMathWalker: It should be populated when https://github.com/actix/actix-extras/blob/1774b8a36ef145bb85e2c0db50c9ff78f844804e/actix-session/src/middleware.rs#L218 is called.

miketwenty1: and this logic needs to be replicated in session.rs? This is tricky for me to understand. This is populated within `get_session_key` fn I'm proposing?

actix-session/src/storage/session_key.rs
miketwenty1: In the code we are impl a `AsRef`
https://github.com/actix/actix-extras/blob/d2b7229f88c5b27c6c8310621e26910fcbcecbfc/actix-session/src/storage/session_key.rs#L38

```
method `as_ref` has an incompatible type for trait
expected fn pointer `fn(&storage::session_key::SessionKey) -> &str`
   found fn pointer `fn(&storage::session_key::SessionKey) -> &secrecy::Secret<std::string::String>`
```

However it's not clear to me how to do with with secrecy data
https://docs.rs/secrecy/0.8.0/secrecy/struct.Secret.html#trait-implementations

Should I be looking for a trait AsRef here?

If I update to
```rust
impl AsRef<secrecy::Secret<String>> for SessionKey {
    fn as_ref(&self) -> &secrecy::Secret<String> {
        &self.0
    }
}
```
Does this mean I need to update redis_rs.rs / redis_actor.rs / cookie.rs

miketwenty1 (commented):
follow to @LukeMathWalker on his feedback for session key exposure. Jan 3, 2023.

miketwenty1:
I'm not familiar with the Github review process, I resolved the comments that looked like duplicates and I pushed a breaking commit so we can discuss/comment on new code.

miketwenty1:
@LukeMathWalker did you need me to do more work on this before getting next bits of feedback? I have some outstanding questions.

PR #320, opened 2023-04-09
rename private-network-access feature

robjtede:
## PR Type

Spec Alignment

## PR Checklist

## Overview
Align with spec changes. 

https://wicg.github.io/private-network-access -> https://wicg.github.io/local-network-access

### DIFF
--- a/.github/workflows/ci-post-merge.yml
+++ b/.github/workflows/ci-post-merge.yml
@@ -1,8 +1,14 @@
 name: CI (post-merge)
 
 on:
-  push:
-    branches: [master]
+  push: { branches: [master] }
+
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
+concurrency:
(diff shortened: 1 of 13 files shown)


latin-1:
I don't believe we should do it right now.

https://github.com/WICG/local-network-access/issues/91#issuecomment-1494704528

> Given that we don't want to discourage developers from adopting security improvements, we believe that it's in users' and developers' best interest to keep the existing header names, `Access-Control-Allow-Private-Network` and `Access-Control-Request-Private-Network`, and we have decided to do so in our implementation. We'll undo https://github.com/WICG/local-network-access/pull/97 to match that reality.

robjtede:
Thanks for surfacing that comment so quickly. I haven't cut a release so will get it reverted soon.

robjtede:
reverted in 8a31f3020e48e4a691524637cdb71496294b5ec8

PR #348, opened 2023-10-03
Add the ability to change the keys

dgagn:
## PR Type

Feature

## PR Checklist

## Overview

**Current Behavior**:
The `IdentityMiddleware` uses hardcoded strings as session keys for user identity and timestamps. These keys are:

- "actix_identity.user_id"
- "actix_identity.last_visited_at"
- "actix_identity.logged_in_at"

**New behavior**:
The `IdentityMiddleware` now provides configuration options to allow users to set custom session keys for user identity and timestamps. These keys can be set via the `IdentityMiddlewareBuilder` and are stored in the `Configuration` struct. The default values remain the same as the original hardcoded strings.

While the default behavior remains the same, users who are extending or interacting directly with the `Configuration` struct or the `IdentityMiddlewareBuilder` might need to adjust their code to accommodate the new fields for custom session keys.

### DIFF
--- a/actix-identity/src/config.rs
+++ b/actix-identity/src/config.rs
@@ -9,6 +9,9 @@ pub(crate) struct Configuration {
     pub(crate) on_logout: LogoutBehaviour,
     pub(crate) login_deadline: Option<Duration>,
     pub(crate) visit_deadline: Option<Duration>,
+    pub(crate) id_key: &'static str,
+    pub(crate) last_visit_unix_timestamp_key: &'static str,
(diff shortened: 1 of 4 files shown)


AprilNEA:
I think this would be very useful for me and hopefully it will get moved forward soon.

LukeMathWalker (approved):

jayvdb (approved):

JohnTitor (approved):
LGTM, thanks!

PR #366, opened 2023-11-19
fix(session): fix double serialisation of values when inserting data into session state

### LINKED ISSUE #308: Actix session insert should check for string
## Expected Behavior

To not store a string as string wrapped in a string. (extra double quotes). Unsure if there's some magic I haven't seen to prevent this.

https://github.com/actix/actix-extras/blob/master/actix-session/src/session.rs#L136-L147
We should check if the value is a string. If it's a string we should not use `serde_json::to_string`
## Current Behavior
Using `serde_json::to_string` on everything including values of type string

## Possible Solution

it seems there may have been uniformity reasons for not allowing types other than strings being stored, but I believe this makes the middleware less useful and more awkward.

1. using redis middleware insert a session value

## Context

Double stringified strings result in all those values needing to be doubly unwrapped which is awkward. Would be nice to let users decide what kind of values they want to store and not wrap everything as string.

## Your Environment

* Rust Version (I.e, output of `rustc -V`):
* Actix-* crate(s) Version:
Latest stable.

zignis:
## PR Type

Bug Fix

## PR Checklist

## Overview

Replaces the usage of HashMap in session with serde Map, which solves the issue of double-serialization of values when inserting data into the session state.

Closes #308


0rangeFox:
Hey 👋. I'm experiencing this issue, can I participate on this PR to be merged asap? At this moment I have a _messy_ workaround and this would simplify the logic and clear my messy code just for this.

PR #380, opened 2024-01-07
feature(settings): add TLS

0rangeFox:
## PR Type
Feature

## PR Checklist

## Overview
Added the logic to get the TLS option working, and separated this component as feature `tls`.


0rangeFox:
The `doc test` workflow is prone to fail due to missing certificates on the [`TLS` example](https://github.com/actix/actix-extras/pull/380/commits/29eec21a0831e321df216b3322df81784a479fc2#diff-1ab1ccecf554146bcf23e8133fc94db9cb0ee2885d7508894961776f84539ec9R22-R49).

actix-settings/src/settings/tls.rs
```diff
@@ -16,3 +17,44 @@ pub struct Tls {
     /// Path to private key `.pem` file.
     pub private_key: PathBuf,
 }
+
+impl Tls {
+    /// Generates an [`SslAcceptorBuilder`] with its settings. It is often used for the following method
+    /// [`actix_web::server::HttpServer::bind_openssl`].
+    ///
+    /// # Example
+    /// ```rust
```
robjtede, line 26:
```suggestion
    /// ```no_run
```

0rangeFox:
Did I miss something while moving the OpenSSL steps from the [`actix-web` workflow](https://github.com/actix/actix-web/blob/891ab083c6d497db2f6adde7767b41e367ca8907/.github/workflows/ci.yml#L38-L45) to this one? It's because from what I can see it still gives an error on the OpenSSL side.

robjtede:
*sigh*... windows and openssl are such great friends... lemme check it out today

actix-settings/src/lib.rs
```diff
@@ -258,14 +260,20 @@ where
...
-            // }
-            unimplemented!("[ApplySettings] TLS support has not been implemented yet.");
-        } else {
-            for Address { host, port } in &settings.hosts {
+        for Address { host, port } in &settings.hosts {
+            #[cfg(feature = "tls")]
+            {
+                if settings.tls.enabled {
+                    self = self.bind_openssl(format!("{}:{}", host, port), settings.tls.get_ssl_acceptor_builder())
+                        .unwrap(/*TODO*/)
```
jayvdb, line 268: some TODOs here and a few lines down.

Perhaps split this into a function which returns a `Result<Self>`, and a deprecated function that returns `Self`

0rangeFox: I couldn't understand exactly what you meant, I think it's eliminating the `unwrap` and then split into functions? Correct me if I'm wrong or maybe I haven't understood your ideology.

jayvdb: I mean revise the existing 

```diff
-fn apply_settings(mut self, settings: &ActixSettings) -> Self {
+fn try_apply_settings(mut self, settings: &ActixSettings) -> Result<Self, Error> {
```

and then append to the trait a new

```rust
#[deprecated]
fn apply_settings(mut self, settings: &ActixSettings) -> Self {
    self.try_apply_settings(settings).unwrap("try_apply_settings failed")
}
```

0rangeFox: Oh, I see, wouldn't that be better done in another PR?

jayvdb: IMO this addition of `.unwrap(/*TODO*/)` warrants it being done in this PR.

previously there was only one .unwrap in this function, and it was for a very predictable and small problem set, typically port already being in use.

In your changes, there are now an additional source of potential panics, and the reasons why SSL init might cause panic are myriad.

0rangeFox: Ahhh ok, in a few hours, I'll take a look at this part and make these changes. Thanks.

jayvdb (approved):

robjtede:
Can the feature be renamed to `openssl` please so we can add `rustls` later.

robjtede (approved):
cool, thanks 👍🏻

PR #381, opened 2024-01-08
feature(session): add deadpool-redis compatibility

0rangeFox:
## PR Type
Feature

## PR Checklist

## Overview
A new feature named `redis-dp-session` has been added, as well as the logic for accepting the [deadpool-redis](https://crates.io/crates/deadpool-redis) library in the [redis-rs](https://github.com/0rangeFox/actix-extras/blob/master/actix-session/src/storage/redis_rs.rs) file, since _deadpool-redis_ uses _redis-rs_ in under the hood, so both are compatible and instead of duplicating, some attributes have been added to control the compilation of the features that are selected.


0rangeFox:
The `doc tests` workflow is failing because there is an [example code](https://github.com/actix/actix-extras/pull/381/commits/b8b2deed42208d82dcc6cf98d0e06cb249a686fb#diff-e5b9dc27fb6f6c0deb6110c827a40a9366b1fb185e6c368fae083ff943a66e9bR67-R74) in `redis_rs` in the `Deadpool Redis` section, but ins't marked only for the feature `redis-dp-session`, is there any way to control this so that the workflow succeeds?

actix-session/src/storage/redis_rs.rs
```diff
@@ -123,21 +165,33 @@ impl RedisSessionStoreBuilder {
...
+    #[cfg(feature = "redis-rs-session")]
+    pub async fn build(self) -> Result<RedisSessionStore, Error> {
         let client = ConnectionManager::new(redis::Client::open(self.connection_string)?).await?;
         Ok(RedisSessionStore {
             configuration: self.configuration,
             client,
         })
     }
+
+    /// Finalise the builder and return a [`RedisActorSessionStore`] instance.
```
jayvdb, line 177: It returns a `RedisSessionStore`

actix-session/src/storage/redis_rs.rs
```diff
@@ -123,21 +165,33 @@ impl RedisSessionStoreBuilder {
...
+    }
 }
 
 impl SessionStore for RedisSessionStore {
     async fn load(&self, session_key: &SessionKey) -> Result<Option<SessionState>, LoadError> {
         let cache_key = (self.configuration.cache_keygen)(session_key.as_ref());
 
         let value: Option<String> = self
-            .execute_command(redis::cmd("GET").arg(&[&cache_key]))
+            .execute_command(cmd("GET").arg(&[&cache_key]))
```
jayvdb, line 194: IMO these changes are undesirable, reducing readability of the code.

actix-session/src/storage/redis_rs.rs
```diff
@@ -259,8 +322,14 @@ impl RedisSessionStore {
...
 
+        #[cfg(feature = "redis-rs-session")]
+        let client = &mut self.client.clone();
+
+        #[cfg(not(feature = "redis-rs-session"))]
+        let client = &mut self.pool.get().await.unwrap();
+
         loop {
-            match cmd.query_async(&mut self.client.clone()).await {
+            match cmd.query_async(client).await {
```
jayvdb, line 332: here we have a `cmd` shadowing the `use ...redis::cmd`

actix-session/src/storage/redis_rs.rs
```diff
@@ -2,7 +2,13 @@ use std::sync::Arc;
 
 use actix_web::cookie::time::Duration;
 use anyhow::Error;
-use redis::{aio::ConnectionManager, AsyncCommands, Cmd, FromRedisValue, RedisResult, Value};
+#[cfg(not(feature = "redis-rs-session"))]
+use deadpool_redis::{
```
jayvdb, line 6: Add
```rust
use deadpool_redis::redis as redis;
```
?

actix-session/src/storage/redis_rs.rs
0rangeFox: I've removed this part because once it's just `redis`, I didn't think there was much need to force it. I'm already reversing that part.

actix-session/src/storage/redis_rs.rs
0rangeFox: That's true, I've already reversed the changes to that part and I've already applied your suggestion of making _aliases_ to the imports.

actix-session/src/storage/redis_rs.rs
0rangeFox: Thanks, I didn't even think of that. I'm already applying all that to avoid certain _shadowing_ and thus also reinforce better reading.

actix-session/src/storage/redis_rs.rs
```diff
@@ -83,31 +101,53 @@ impl RedisSessionStore {
...
+    }
+
     /// Create a new instance of [`RedisSessionStore`] using the default configuration.
     /// It takes as input the only required input to create a new instance of [`RedisSessionStore`] - a
     /// connection string for Redis.
-    pub async fn new<S: Into<String>>(
-        connection_string: S,
-    ) -> Result<RedisSessionStore, anyhow::Error> {
+    #[cfg(feature = "redis-session")]
+    pub async fn new<S: Into<String>>(connection_string: S) -> Result<RedisSessionStore, Error> {
```
robjtede, line 127: i'd rather not introduce a crate feature that makes things non-additive.

0rangeFox: Sorry, but I didn't understand what you meant by that, what do you mean? About the new `deadpool-redis` feature or...?

robjtede:
This is actually blocked on a new release of deadpool-redis with a redis v0.26 dependency update.

actix-session/Cargo.toml
```diff
@@ -39,6 +40,9 @@ tracing = { version = "0.1.30", default-features = false, features = ["log"] }
 # redis-session
 redis = { version = "0.26", default-features = false, features = ["tokio-comp", "connection-manager"], optional = true }
 
+# redis-dp-session
+deadpool-redis = { version = "0.14", optional = true }
```
rgrfoss, line 44: version = "0.16.0"

robjtede (approved):

robjtede:
thanks for your work on this @0rangeFox

PR #395, opened 2024-01-31
doc(actix-cors): update MSRV

marvin-j97:
I got the error message:
```
error: package `actix-cors v0.7.0` cannot be built because it requires rustc 1.75 or newer, while the currently active rustc version is 1.74.1
```

actix-cors [uses the workspace's rust-version](https://github.com/actix/actix-extras/blob/master/actix-cors/Cargo.toml#L14), [which is 1.75](https://github.com/actix/actix-extras/blob/master/Cargo.toml#L19)


robjtede (approved):
thanks 👍🏻

PR #407, opened 2024-02-28
chore: update futures & futures-core to 0.3.30 across workspace

zopieux:
## PR Type

Other

## PR Checklist

- [x] Tests for the changes have been added / updated. (not applicable)
- [x] Documentation comments have been added / updated. (not applicable)
- [x] A changelog entry has been made for the appropriate packages. (not applicable)
- [x] Format code with the nightly rustfmt (`cargo +nightly fmt`). (not applicable)

## Overview

The ancient `futures` & `futures-core` versions currently pinned are causing obscure issues with e.g. `Identity` (in actix-identity) not being recognized anymore by just adding the `futures` crate (any version) in a project that also depends on actix-identity. Bumping `futures` version in actix-* fixes it.

### DIFF
--- a/actix-identity/Cargo.toml
+++ b/actix-identity/Cargo.toml
@@ -24,7 +24,7 @@ actix-utils = "3"
 actix-web = { version = "4", default-features = false, features = ["cookies", "secure-cookies"] }
 
 derive_more = "0.99.7"
-futures-core = "0.3.7"
+futures-core = "0.3.30"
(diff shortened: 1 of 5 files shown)


robjtede:
Thank you for the report and contribution.

Increasing these version requirements is rarely the true solution to problems such as what you're describing. I've aligned the requirements to `^0.3.17` but without a reproduction case I don't want to accept this PR.

PR #467, opened 2024-09-28
Update websocket example to use futures_util::StreamExt

MatthewSbar:
The only way I was able to get this example code to run was by installing futures_util and importing futures_util::StreamExt. It feels like there isn't supposed to be an additional dependency here (I am relatively new to Rust let alone package management like this), but the example by itself throws an error without doing this.

## PR Type

Other

## PR Checklist

## Overview

It's just an update to the example code

N/A to my knowledge

### DIFF
--- a/actix-ws/README.md
+++ b/actix-ws/README.md
@@ -20,6 +20,7 @@
 ```rust
 use actix_web::{middleware::Logger, web, App, HttpRequest, HttpServer, Responder};
 use actix_ws::Message;
+use futures_util::StreamExt;
 
 async fn ws(req: HttpRequest, body: web::Payload) -> actix_web::Result<impl Responder> {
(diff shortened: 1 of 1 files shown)


MatthewSbar:
I recognize this probably isn't the correct-est solution, but I found merely pasting the example into a main.rs file and trying to do `cargo run` would give me this error, importing the futures_util package gave me the expected behavior. Maybe I'm doing something wrong 🙃, but, I wanted to try and do something more constructive than say "it doesn't work" 

```
    |
115 |         while let Some(Ok(msg)) = msg_stream.next().await {
    |                                              ^^^^ method not found in `MessageStream`
    |
```

MatthewSbar:
Oh looks like a fix was reported an hour ago and merged a few minutes ago. Closing.

robjtede:
Yep sorry about this! Appreciate the PR regardless :)

PR #468, opened 2024-09-29
Upstream `tracing-actix-web` into `actix-extras`

LukeMathWalker:


LukeMathWalker:
There's a few failing jobs:

- All the MSRV ones fail because `tracing-actix-web` needs Rust 1.80 because of `LazyLock`
- Public API diff fails because there's no `tracing-actix-web` in `main`, as far as I can see

Let me know how you want to move forward on the first one.

JohnTitor:
@LukeMathWalker I've merged the repo as subtree in #615!
Will you take the future release responsibility or want the team to do so? If latter, it'd be great if you could add me to the crate owners, when you have some time 🙏

PR #473, opened 2024-10-02
Include cookie expiration and ttl in the identity example

jwiesler:
## PR Type

Other/Improve examples

## PR Checklist

- [x] Tests for the changes have been added / updated. (does not apply

## Overview
I ran into this issue that on some mobile browsers not setting a ttl of the session cookie will actually delete the cookie on session end (some time when the app goes into sleep mode).
I noticed this issue for almost two years now but could never figure out why it was happening. It is very hard to diagnose since it is sporadic and only happens on some browsers (here it was Brave).

I think it's very helpful to include `session_ttl` in the example, I guess many usages of the identity middleware are to enable login flows with some kind of expiration and I just did not know about this at all.

This only changes the example of actix-identity, I ran it locally and it works as expected.

### DIFF
--- a/actix-identity/examples/identity.rs
+++ b/actix-identity/examples/identity.rs
@@ -13,10 +13,10 @@
 //! http -v --session=identity GET localhost:8080/
 //! ```
 
-use std::io;
+use std::{io, time::Duration};
 
 use actix_identity::{Identity, IdentityMiddleware};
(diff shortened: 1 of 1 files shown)


robjtede (approved):
nice tweak, thanks 👍🏻

PR #486, opened 2024-12-15
feat: added PartialEq to Cors

### LINKED ISSUE #302: Allow testing created `Cors` instance
Creating a `Cors` middleware requires configuration in some cases. That needs testing.

However, it is pretty hard to test a created `Cors` instance as it doesn't expose its values.

## Expected Behavior

Be able to test a `Cors` instance for its values.

## Current Behavior

Not possible. It is possible to render a debug string of the struct, however as it uses `HashSet` with `RandomStates`, instances containing `HashSet` values can't be tested.

## Possible Solution

Have `Cors` implement `PartialEq`. Or allow access to the values of its `Inner`.

1. Set up a `Cors` instance
2. Try to test it.
3.
4.

## Your Environment

* Rust Version (I.e, output of `rustc -V`):
* Actix-* crate(s) Version:

CommanderStorm:
## PR Type

Feature

## PR Checklist

## Overview

I have added `impl PartialEq for Cors` which should allow the testability improvements asked in #302

No breaking changes as far as I can tell

Closes #302


actix-cors/src/inner.rs
```diff
@@ -27,6 +27,12 @@ impl Default for OriginFn {
     }
 }
 
+impl PartialEq for OriginFn {
+    fn eq(&self, other: &Self) -> bool {
+        Rc::ptr_eq(&self.boxed_fn, &other.boxed_fn)
```
CommanderStorm, line 32: If this equality here is correct is something that needs debating.
Is function equivalence for you:
- (current) literal => i.e. the pointers need to match
- semantic => i.e. they have to do the same thing

(If it should be semantic, we can close this PR and the linked issue, as I don't think that this is possible)

CommanderStorm:
(The CI fail seems unrelated as it also fails in https://github.com/actix/actix-extras/pull/480)

actix-cors/src/inner.rs
robjtede: point equality seems fine here

robjtede (approved):

PR #496, opened 2025-01-16
Improve logout example

wt:
The current example for logout is not show a complete example.

I have added a couple lines to handle the case where a logged out user tries to logout again.

## PR Type

Bug Fix


robjtede (approved):
thanks 👍🏻

PR #585, opened 2025-09-09
[NOJIRA-000] Extract http headers for actix-protobuf

jlloh:
## PR Type

Feature

## PR Checklist

## Overview

Extract HeaderMap and pass it along so it can be used in the handler.


JohnTitor (changes requested):
Thanks for the PR. I, however, would like to reject this PR as making `ProtoBuf<T>` always carry a `HeaderMap` forces every existing handler to break, with no alternative other than fabricating an empty map. Headers should be configurable/extractable in the request.

PR #724, opened 2026-04-30
fix(tracing-actix-web): populate trace_id at span creation

tyrypyrking:
## PR Type

Bug Fix

## PR Checklist

## Overview

When a request arrives with a `traceparent` header, the `trace_id` was missing from the "REQUEST START" log entry and only appeared at "REQUEST END". The cause: `trace_id` was set via `span.record()` *after* the span had already been created, but subscribers like `tracing-bunyan-formatter` emit START during `on_new_span` — at which point the field is still `Empty`.

This PR extracts the `trace_id` from the propagated context *before* creating the span, so it is baked into the span fields from the start. When there is no remote parent (no `traceparent` header), the field stays `Empty` and gets filled later by `set_otel_parent` exactly as before — no behavior change in that path.

### What changed

- **`tracing-actix-web/src/otel.rs`** — new `extract_trace_id()` that pulls the trace_id from request headers via the global propagator.
- **`tracing-actix-web/src/root_span_macro.rs`** — `root_span!` now calls `extract_otel_trace_id()` before span creation; two `inner_span` arms handle the "known trace_id" vs "empty" cases (unavoidable since `tracing::Value` is sealed).

No breaking changes.

Original PR was at the old repo: https://github.com/LukeMathWalker/tracing-actix-web/pull/145

### DIFF
--- a/tracing-actix-web/CHANGELOG.md
+++ b/tracing-actix-web/CHANGELOG.md
@@ -3,6 +3,9 @@
 ## Unreleased
 
 - Minimum supported Rust version (MSRV) is now 1.88.
+- Ensure the OpenTelemetry `trace_id` is populated when request spans are created. [#724]
+
+[#724]: https://github.com/actix/actix-extras/pull/724
 
 ## 0.7.21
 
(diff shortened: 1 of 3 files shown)


JohnTitor (approved):
Thanks! LGTM.

PR #765, opened 2026-06-20
refactor(tracing): decouple OTel support into adapter crate

JohnTitor:
## PR Type

Refactor 

## PR Checklist

## Overview

This is an attempt to remove the OTel features to support. Currently it becomes harder and harder to maintain, and old versions should be quite lower usage than newer's.
This decouples the part of code which implements OTel support as another adapter crate so that we can versioning them independently. The new adapter crate matches the OTel crate version so users can easily expect what version they should use.
And the main crate can contain any functional changes independently so users can find combination they want (e.g. users can use the latest core crate feature + older OTel).

cc @LukeMathWalker this is kinda drastic change and would love to hear your inputs if you have!


tracing-actix-web/examples/custom-root-span/src/main.rs
JohnTitor, line 1: OTel users would have to migrate like this example. I guess it's not that painful.

tracing-actix-web/src/trace_context.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+/// Integrates a distributed tracing backend with [`TracingLogger`].
+///
+/// `tracing-actix-web` itself does not depend on a tracing backend. Implement
+/// this trait in an adapter crate to extract remote trace context from request
+/// headers and attach it to the root [`tracing::Span`].
+///
+/// [`TracingLogger`]: crate::TracingLogger
+pub trait TraceContext: Clone + 'static {
```
JohnTitor, line 21: This is for adapter crate. One bonus point, users can alternatively implement their desired backend if want.

tracing-actix-web-opentelemetry/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "tracing-actix-web-opentelemetry"
+version = "0.32.0"
```
JohnTitor, line 3: I'm going to match the version here with `opentelemetry`.

.github/workflows/lint.yml
```diff
@@ -66,17 +66,10 @@ jobs:
           reporter: github-pr-check
           github_token: ${{ secrets.GITHUB_TOKEN }}
           clippy_flags: >-
-            --workspace --all-features --exclude actix-settings --exclude tracing-actix-web
+            --workspace --all-features --exclude actix-settings
             --tests --examples --bins --
             -A unknown_lints -D clippy::todo -D clippy::dbg_macro
 
-      - name: Check tracing-actix-web with Clippy
```
JohnTitor, line 85: This is something I want to remove by this change :>

PR #777, opened 2026-07-10
Be more precse about Cors::permissive() about HTTP methods

sseering:
## PR Type

Other 

## PR Checklist
- [✓] Documentation comments have been added / updated.

## Overview

Be more precse about Cors::permissive() about HTTP methods. The old documentation suggest it accepts all HTTP methods (as in: any string). But it doesn't accept the new QUERY HTTP method. The new method isn't accepted as upstream hasn't added it yet.

### DIFF
--- a/actix-cors/src/builder.rs
+++ b/actix-cors/src/builder.rs
@@ -92,7 +92,7 @@ impl Cors {
     /// Constructs a very permissive set of defaults for quick development. (Not recommended for
     /// production use.)
     ///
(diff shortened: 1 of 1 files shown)


robjtede:
> doesn't accept the new QUERY HTTP method

we can add it to the `ALL_METHODS_SET` list, it doesn't need the constant on Method to be used here

JohnTitor:
Actually we have a PR for it: https://github.com/actix/actix-web/pull/4140
Closing as this will be resolved in the near future by it.

